PR #2, opened 2024-10-07
Initial stage of charm rewrite

yhaliaw:
Applicable spec: ISD-183

### Overview

This implements the MVP for the content-cache charm.

The charm includes:
- Getting the configurations in integration data
- Inject the configuration to nginx config files, and manage them.
- Manages the nginx instance.

### Rationale

This is the initial code for the charm.

### Juju Events Changes

Added the following events:
- start: Initialize the nginx. Set the unit status.
- stop: Stop the nginx.
- update-status: Set the unit status.
- relation-changed: Validate the integration data and store it as nginx configuration file (overwriting the old one). Load the config.
- relation-broken:  Validate the integration data and store it as nginx configuration file (overwriting the old one). Load the config. This is to remove the config from the broken config.

### Module Changes

Added the state.py module for the validation of configurations from integration data.

Added the nginx_manager.py module for management of nginx and the nginx configuration files.

Added the error.py module for storing the error classes.

### Checklist

- [ ] The documentation for charmhub is updated


.licenserc.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - '.jujuignore'
+    - '.gitignore'
+    - '.licenserc.yaml'
+    - 'CODEOWNERS'
+    - 'icon.svg'
+    - 'LICENSE'
+    - '.pylintrc'
+    - '.woke.yaml'
+    - 'lib/**'
+    - 'tests/integration/data/clouds.yaml.tmpl'
```
gregory-schiano, line 28: Probably a left over

charmcraft.yaml
```diff
@@ -0,0 +1,18 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+type: charm
+
+bases:
+  - build-on:
+    - name: ubuntu
+      channel: "22.04"
```
gregory-schiano, line 9: Shouldn't it be 24.04 ?

metadata.yaml
```diff
@@ -0,0 +1,34 @@
...
+source: https://github.com/canonical/is-charms-template-repo
+
+description: |
+  A single sentence that says what the charm is, concisely and memorably.
+
+  A paragraph of one to three short sentences, that describe what the charm does.
+
+  A third paragraph that explains what need the charm meets.
+
+  Finally, a paragraph that describes whom the charm is useful for.
```
gregory-schiano, line 30: Should be filled in

pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+skips = ["*/*test.py", "*/test_*.py", "*tests/*.py"]
+
+# Testing tools configuration
+[tool.coverage.run]
+branch = true
+
+# Formatting tools configuration
+[tool.black]
+line-length = 99
+target-version = ["py38"]
```
gregory-schiano, line 16: If we're on noble we could target 3.12

pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+max-line-length = 99
+max-doc-length = 99
+max-complexity = 10
+exclude = [".git", "__pycache__", ".tox", "build", "dist", "*.egg_info", "venv"]
+select = ["E", "W", "F", "C", "N", "R", "D", "H"]
+# Ignore W503, E501 because using black creates errors with this
+# Ignore D107 Missing docstring in __init__
+ignore = ["W503", "E501", "D107"]
+# D100, D101, D102, D103: Ignore missing docstrings in tests
+per-file-ignores = ["tests/*:D100,D101,D102,D103,D104,D205,D212,D415"]
```
gregory-schiano, line 32: Check why D104, D205, D212 and D415

requirements.txt
```diff
@@ -0,0 +1,3 @@
+ops == 2.17.0
+pydantic == 2.9.2
+python-nginx ~= 1.5
```
gregory-schiano, line 3: Should be pinned to the exact version

src/charm.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def _set_status(self) -> None:
+        """Set the charm status."""
+        if not self.model.relations[CACHE_CONFIG_INTEGRATION_NAME]:
+            self.unit.status = ops.BlockedStatus(WAIT_FOR_CONFIG_MESSAGE)
+            return
+
+        if not self._nginx.ready_check():
+            self.unit.status = ops.MaintenanceStatus(NGINX_NOT_READY_MESSAGE)
+        else:
```
gregory-schiano, line 74: `return` and no need for else

src/charm.py
```diff
@@ -0,0 +1,95 @@
...
+        except IntegrationDataError as err:
+            self.unit.status = ops.BlockedStatus(str(err))
+            return
+        if not nginx_config:
+            self._nginx.stop()
+            self.unit.status = ops.BlockedStatus(WAIT_FOR_CONFIG_MESSAGE)
+            return
+
+        self._nginx.update_config(nginx_config)
+        self._nginx.load()
```
gregory-schiano, line 90: `load` could be confusing

src/nginx_manager.py
```diff
@@ -0,0 +1,75 @@
...
+
+NGINX_SITES_ENABLED_PATH = Path("/etc/nginx/sites-enabled")
+NGINX_DEFAULT_SITES_PATH = NGINX_SITES_ENABLED_PATH / "default"
+CHARM_MANAGED_NGINX_CONFIG_PATH = NGINX_SITES_ENABLED_PATH / "charm-managed"
+
+
+# Unit test is not valuable as the class is closely coupled with nginx.
+# This should be tested with integration tests.
+class NginxManager:  # pragma: no cover
+    """Manages the Nginx instance."""
```
gregory-schiano, line 21: Don't think a class is necessary, it can only a module

src/nginx_manager.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def init(self) -> None:
+        """Initialize the nginx server."""
+        # The install, systemctl enable, and systemctl start are idempotent.
+        execute_command(["sudo", "apt", "install", "nginx", "-yq"])
+        execute_command(["sudo", "systemctl", "enable", "nginx"])
+        execute_command(["sudo", "systemctl", "start", "nginx"])
```
gregory-schiano, line 28: We should probably handle failures here

src/nginx_manager.py
```diff
@@ -0,0 +1,75 @@
...
+        """Start or reload the nginx server."""
+        if self.ready_check():
+            # This is reload the configuration files without interrupting service.
+            execute_command(["sudo", "nginx", "-s", "reload"])
+        else:
+            execute_command(["sudo", "systemctl", "restart", "nginx"])
+
+    def stop(self) -> None:
+        """Stop the nginx server."""
+        execute_command(["sudo", "nginx", "-s", "quit"])
```
gregory-schiano, line 42: Why not using systemctl stop ?

src/state.py
```diff
@@ -0,0 +1,128 @@
...
+        try:
+            relation_data = rel.data[rel.app]
+            if not relation_data:
+                logger.info("Found integration %s with no data", rel.id)
+                continue
+            config = LocationConfig.from_integration_data(relation_data)
+        except ConfigurationError as err:
+            raise IntegrationDataError(
+                f"Faulty data from integration {rel.id}: {str(err)}"
+            ) from err
```
gregory-schiano, line 126: We should keep processing if only 1 relation has issue, to avoid stopping serving the service for the other relation

src/nginx_manager.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def update_config(self, configuration: NginxConfig) -> None:
+        """Update the nginx configuration files.
+
+        Args:
+            configuration: The nginx locations configurations.
+        """
+        nginx_config = nginx.Conf()
+        server_config = nginx.Server()
+        for i, config in enumerate(configuration.values()):
```
gregory-schiano, line 52: You're iterating over locations, so probably better to name it "location_config", or location

src/state.py
```diff
@@ -0,0 +1,128 @@
...
+        Args:
+            data: One set of integration data.
+
+        Raises:
+            ConfigurationError: Invalid cache configurations in integration data.
+
+        Returns:
+            The object.
+        """
+        location = data.get(LOCATION_CONFIG_NAME, "").strip()
```
gregory-schiano, line 64: Beware that this location is used in the nginx `location` attribute, that only supports path. We should parse it

src/nginx_manager.py
```diff
@@ -0,0 +1,75 @@
...
+            upstream_config = nginx.Upstream(backend_set_name, *hosts)
+            nginx_config.add(upstream_config)
+            server_config.add(
+                nginx.Location(
+                    config.location,
+                    nginx.Key("proxy_pass", f"{config.protocol}://{backend_set_name}"),
+                )
+            )
+        nginx_config.add(server_config)
+        nginx.dumpf(nginx_config, CHARM_MANAGED_NGINX_CONFIG_PATH)
```
gregory-schiano, line 66: Should probably do some error handling here

gregory-schiano: You should consider file permission / ownership

gregory-schiano (commented):
Did a first peer review

pyproject.toml
yhaliaw: I removed D104.
These ignore are for tests (`test/*`).
The rest conflicts with the arrange, act, assert with tests. Added a comment on that.

metadata.yaml
```diff
@@ -13,21 +13,19 @@ name: content-cache
...
 description: |
-  A single sentence that says what the charm is, concisely and memorably.
+  A [Juju](https://juju.is/) [charm](https://juju.is/docs/olm/charmed-operators) deploying and managing a static web content cache with nginx.
 
-  A paragraph of one to three short sentences, that describe what the charm does.
-
-  A third paragraph that explains what need the charm meets.
-
-  Finally, a paragraph that describes whom the charm is useful for.
+  This machine charm manages a nginx instance configured as a content cache.
```
gregory-schiano, line 26: We should probably explain a bit what we mean by "Content cache"

src/nginx_manager.py
```diff
@@ -3,73 +3,238 @@
...
+        logger.exception(
+            "Unable to convert %s configuration to nginx format: %s", host, configuration
+        )
+        raise NginxConfigurationError(
+            f"Unable to convert {host} configuration to nginx format: {configuration}"
+        ) from err
+
+    try:
+        nginx.dumpf(nginx_config, _get_sites_available_path(host))
+        _get_sites_enabled_path(host).symlink_to(_get_sites_available_path(host))
```
gregory-schiano, line 186: Perhaps add a small comment to explain why you're doing this, somebody less familiar with nginx might not understand

gregory-schiano: Also both _get_sites_enabled_path and _get_sites_available_path are only used here and are 1 liner methods
I'd rather have a method called `enable_site(host: str)`
That will build the path, and create the symlink, then no need for a comment because the method will be documented via docstrings

yhaliaw: I added a function that creates the nginx config file and add symlink, with the docstring explaining why.

I did tried to delete the `_get_sites_enabled_path` and `_get_sites_available_path` functions, but I think it might be better to keep them as they are used in tests. It is likely better for the test to use the same function to get the path as the charm to reduce code duplicate and regressions.

metadata.yaml
yhaliaw: Updated.

gregory-schiano (approved):

src/charm.py
```diff
@@ -0,0 +1,145 @@
...
+            )
+            raise
+        except NginxConfigurationAggregateError as err:
+            logger.exception("Found error with configuration for hosts: %s", err.hosts)
+            logger.warning(
+                "Any hosts configuration without errors will be served on content cache"
+            )
+            status_message = f"Error for host: {err.hosts}"
+
+        nginx_manager.load_config()
```
jdkandersson, line 114: I was thinking about the interface, couldn't the update and load config be the same function? What would be the advantages of separating them?

src/nginx_manager.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+def load_config() -> None:  # pragma: no cover
+    """Load nginx configurations."""
+    if ready_check():
+        logger.info("Loading nginx configuration files")
+        # This is reload the configuration files without interrupting service.
+        execute_command(["sudo", "nginx", "-s", "reload"])
+        return
+    logger.info("Restarting nginx to load the configuration files.")
```
jdkandersson, line 63: nitpick: I would leave an empty line here to indicate that these are different code blocks

src/nginx_manager.py
```diff
@@ -0,0 +1,260 @@
...
+    """Update the nginx configuration files.
+
+    Raises:
+        NginxConfigurationAggregateError: All failures related to creating nginx configuration.
+        NginxFileError: File operation errors while updating nginx configuration files.
+
+    Args:
+        configuration: The nginx locations configurations.
+    """
+    # This will reset the file permissions.
```
jdkandersson, line 100: Is that all the function does? Just based on the function name I was guessing it would remove customisations in the file and convert it back to a default/ blank configuration file

tests/integration/test_charm.py
```diff
@@ -0,0 +1,106 @@
...
+@pytest.mark.abort_on_fail
+@pytest.mark.asyncio
+async def test_charm_integrate_with_no_data(
+    app: Application, config_app: Application, model: Model
+) -> None:
+    """
+    arrange: A working application of content-cache charm, with no integrations.
+    act:
+        1. Integrate with a charm that provides configuration.
+        2. Add configuration to the charm that provides data.
```
jdkandersson, line 35: I'm not following what these are? Which one is the integrator charm please?

src/charm.py
yhaliaw: There is a bit of discussion here: https://github.com/canonical/content-cache-operator/pull/2#discussion_r1796552668

docs/index.md
```diff
@@ -0,0 +1,36 @@
...
+<!-- A single sentence that says what the product is, succinctly and memorably. Add a 1-2 sentence description of what the charm software does. -->
+
+A [Juju](https://juju.is/) [charm](https://juju.is/docs/olm/charmed-operators) deploying and managing a static web content cache with nginx.
+
+This machine charm manages a nginx instance configured as a content cache. The configuration for the locations of cache is managed with the [Content Cache Backends Config subordinate charm](https://charmhub.io/content-cache-backends-config). Each Content Cache Backends Config charm stores the configuration for a location and the associated set of backends.
+
+<!-- A paragraph of 2-5 short sentences, that describes what the product does and what need the product meets. -->
+
+<!-- Finally, a paragraph that describes whom the product is useful for. -->
+
```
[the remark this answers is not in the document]
yhaliaw: Done.

src/nginx_manager.py
yhaliaw: The function will ensure the configuration file directories are created, empty, and has the right permissions.

I removed the code comment on line 100 as it might be unnecessary.

tests/integration/test_charm.py
yhaliaw: I changed the terminology to `configuration charm` in all the docstrings and code comments of the file.

src/charm.py
jdkandersson: Hmm, what would be the issue of having a function `update_and_load_config(config)`? Would we ever update the config without the immediately loading it?

yhaliaw: I do not think of any case the would involve update and load config separately.

However, I feel splitting off the `load_config` in its own function might be fine, since the update config part will get more complex. For the next PR I am working on, I have split of update configuration to different single-usage functions due to the linting complaining this function is too complex.

tests/unit/test_nginx_manager.py
```diff
@@ -49,12 +50,13 @@ def test_reset_file_with_existing_files(patch_nginx_manager_path: None):
...
 
 
-def test_update_config_with_valid_config(patch_nginx_manager_path: None):
+def test_update_config_with_valid_config(monkeypatch, patch_nginx_manager_path: None):
     """
     arrange: Valid configuration data.
     act: Create configuration files from the data.
     assert: The files are created and has the configurations.
     """
+    monkeypatch.setattr("nginx_manager._load_config", MagicMock())
```
gregory-schiano, line 59: `_load_config` has 2 cases, one where it reloads, and one where it restarts
Shouldn't we have a unittest that cover both cases instead of mocking the whole method ?

yhaliaw: I will fix this in a follow-up PR.

gregory-schiano (approved):

PR #11, opened 2024-12-04
Enable active healthchecks

seb4stien:
Applicable specs:
- [ISD178: Overview of content-cache charm rewrite](https://docs.google.com/document/d/1I3HWYlmX4MoJzLLZJaQr0W18goWOFb-53uZLxYXBblM/edit?tab=t.0)
- [ISD183: implementation for content-cache charm](https://docs.google.com/document/d/1tlpnOOuo42dUAWxa1cAs8U1UROjM6Xv_QCKV9zuN70k/edit?tab=t.0)

### Overview

On the content-cache side, this PR add support for active healthchecks through https://github.com/openresty/lua-upstream-nginx-module.

On the content-cache-backends-config side, this PR enables the configuration of the healthchecks through two variables:
- healthcheck-path
- healthcheck-interval

### Rationale

By default, Nginx only support passive healthchecks. For highly visible and requested websites, it may lead to a bad user experience if a backend goes down. By providing active healthchecks, we aim at limiting the impact of a backend outage.

### Juju Events Changes

Main changes happen in nginx_manager.initialize which is called "on-start" to create the configurations for nginx and later on "relation-changed".

#### state.py

Now supports the two configuration variables healthcheck-path and heathcheck-interval.

#### nginx_manager.py

It now generates the following configurations:
- modules-enabled/lua_upstream.conf: to load the required lua modules
- conf.d/healthcheck.conf: global configuration for the healthchecks
- sites-enables/nginx_backends_status: to get the status of the backends

It extend current virtualhost configuration with a LUA script to init a worker in charge of monitoring the backends.

### Library Changes

None

### Checklist

- [ ] The documentation is generated using `src-docs`
- [ ] The documentation for charmhub is updated
- [ ] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)
- [ ] The docs/changelog.md is updated with user-relevant changes. 

I'll update the doc after a first review, to see if we keep the current approach.


content-cache-backends-config/src/state.py
```diff
@@ -126,8 +130,96 @@ class Configuration(pydantic.BaseModel):
...
+        """Validate the healthcheck_path.
+
+        Args:
+            value: The value to validate.
+
+        Returns:
+            The value after validation.
+        """
+        return _validate_path_value(value)
+
```
yhaliaw, line 222: I think line 141 to 222 should be removed.

The code was refactored to use `pydantic.AfterValidator`. E.g., for the `hostname`:
```
hostname: typing.Annotated[
        str,
        pydantic.StringConstraints(min_length=1),
        pydantic.AfterValidator(_validate_hostname_value),
    ]
```

I think due to merge conflict this part is being added back.

content-cache/src/nginx_manager.py
```diff
@@ -65,16 +95,40 @@ def initialize() -> None:  # pragma: no cover
     """
     logger.info("Installing and enabling nginx")
     # The install, systemctl enable, and systemctl start are idempotent.
-    return_code, _, stderr = execute_command(["sudo", "apt", "install", "nginx", "-yq"])
+    return_code, _, stderr = execute_command(
+        ["sudo", "apt", "install", "nginx", "lua-resty-core", "-yq"]
```
yhaliaw, line 99: For the `lua-resty-core`, is using this APT package the long term plan?

content-cache/src/state.py
```diff
@@ -181,6 +191,8 @@ def from_integration_data(cls, data: ops.RelationDataContent) -> "LocationConfig
         backends_str = data.get(BACKENDS_FIELD_NAME, "").strip()
         fail_timeout = data.get(FAIL_TIMEOUT_FIELD_NAME, "").strip()
         backends_path = data.get(BACKENDS_PATH_FIELD_NAME, "").strip()
+        healthcheck_path = data.get(HEALTHCHECK_PATH_FIELD_NAME, "").strip()
+        healthcheck_interval = int(data.get(HEALTHCHECK_INTERVAL_FIELD_NAME, "2000").strip())
```
yhaliaw, line 195: It is a bit unclear why 2000 is chosen as the default.
Is there a special reason for it?

yhaliaw (commented):
Reminder to updated the `docs/changelog.md` in both of the charm.

content-cache/src/nginx_manager.py
```diff
@@ -299,6 +407,40 @@ def _get_upstream_config_keys(config: LocationConfig) -> tuple[nginx.Key, ...]:
...
+            fall = 3,
+            rise = 2,
+            valid_statuses = {{200}},
+            concurrency = 10,
+        }}
+        if not ok then
+            ngx.log(ngx.ERR, "failed to spawn health checker: ", err)
+            return
+        end
+    """
```
jdkandersson, line 441: Perhaps you want to use https://docs.python.org/3.6/library/textwrap.html#textwrap.dedent so that it is nicely indented when written to a file

content-cache/tests/integration/test_healthchecks.py
```diff
@@ -0,0 +1,192 @@
...
+    assert response.status_code == 200
+    assert http_ok_message in response.content.decode("utf-8")
+
+    # Sample status page
+    #
+    # Nginx Worker PID: 7905
+    # Upstream 88c26973-5726-4745-ab4a-d3addea80d82
+    # Primary Peers
+    #    10.14.1.77:80 UP
+    # Backup Peers
```
jdkandersson, line 92: Can this be removed now? We would generally prefer not to have commented out code please

.github/workflows/integration_test.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
     with:
       juju-channel: 3.6/stable
       provider: lxd
-      modules: '["test_basic", "test_tls_cert", "test_metric"]'
+      modules: '["test_charm", "test_tls_cert", "test_healthchecks", "test_metric"]'
```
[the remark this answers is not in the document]
seb4stien: Thanks for spotting it. Fixed in 1f6ef5f1a8e862aaade3004d5da0ef0ae13b8c01.

content-cache/src/nginx_manager.py
seb4stien: As its part of the distribution, would there be a reason not to use it?

content-cache/src/state.py
seb4stien: I just reused the value from the example provided in the documentation from the module. https://github.com/openresty/lua-resty-upstream-healthcheck
This is something we can change for sure. Or at least put it as a constant maybe?

content-cache/tests/integration/test_healthchecks.py
seb4stien: We can, but don't we create some additional cognitive load to someone who would take over the code in the future?

content-cache/src/nginx_manager.py
```diff
@@ -57,6 +64,29 @@
 # This should be tested with integration tests.
 
 
+class NginxLuaSection:  # pylint: disable=R0903
```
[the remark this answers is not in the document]
seb4stien: Sure. I'll do. First I wanted to clarify if this the way we want to go (introducing such a class to handle nginx configuration).

content-cache/src/nginx_manager.py
seb4stien: Good point.
As for the NginxLuaSection class: I wanted first to get feedback on the approach toward LUA code => is it the way we want to go.

content-cache/src/state.py
```diff
@@ -181,6 +191,8 @@ def from_integration_data(cls, data: ops.RelationDataContent) -> "LocationConfig
         backends_str = data.get(BACKENDS_FIELD_NAME, "").strip()
         fail_timeout = data.get(FAIL_TIMEOUT_FIELD_NAME, "").strip()
         backends_path = data.get(BACKENDS_PATH_FIELD_NAME, "").strip()
+        healthcheck_path = data.get(HEALTHCHECK_PATH_FIELD_NAME, "").strip()
```
gregory-schiano, line 194: Same comment as for `content-cache-backends-config` default value

gregory-schiano (commented):
Didn't review the tests changes to content-cache

content-cache/src/state.py
yhaliaw: I think we should use a good default value for the charm.
The current default of 2000 works out to 2 seconds, which seems a bit too frequent to me.
What is the default value used in the package? Maybe set it to that. 

If the value needs to be updated then remember to update the config file as well: 
`content-cache-backends-config/config.yaml`

seb4stien: The module itself has no default.
I was looking at HA-proxy, and it has a similar default of 2s: https://www.haproxy.com/documentation/haproxy-configuration-tutorials/service-reliability/health-checks/#change-the-interval
```
Change the interval 
By default, the load balancer sends a health check every two seconds. Change this by adding the inter argument to the server line. In the next example, we send a health check every four seconds:

```

As the goal of active healthchecks is to limit impact on end-users, it makes sense to have a small interval. For a static healthcheck endpoint, 1req/2s is not generating any load.

content-cache-backends-config/src/state.py
```diff
@@ -173,6 +265,12 @@ def from_charm(cls, charm: ops.CharmBase) -> "Configuration":
             raise ConfigurationError("Empty backends configuration found")
         fail_timeout = typing.cast(str, charm.config.get(FAIL_TIMEOUT_CONFIG_NAME, "")).strip()
         backends_path = typing.cast(str, charm.config.get(BACKENDS_PATH_CONFIG_NAME, "")).strip()
+        healthcheck_path = typing.cast(
+            str, charm.config.get(HEALTHCHECK_PATH_CONFIG_NAME, "")
```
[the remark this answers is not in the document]
seb4stien: I applied the same strategy as "path" and "backend_path".
What should we align on?
=> put "real" default values for everything in the state?
=> or don't put any default values?

I don't have enough experience with the "states", but I feel duplicating the information is not good, so I would rather go for the second option (and remove the default value I have put for healthcheck_interval).

content-cache/charmcraft.yaml
```diff
@@ -11,6 +11,48 @@ platforms:
...
+      - libxslt1-dev
+      - zlib1g-dev
+    override-build: |
+      set -eux
+      # Get NGINX source
+      mkdir /tmp/nginx-src
+      cd /tmp/nginx-src
+      wget http://nginx.org/download/nginx-1.24.0.tar.gz
+      tar -xf nginx-1.24.0.tar.gz
+      cd nginx-1.24.0
```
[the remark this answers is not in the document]
seb4stien: Indeed. I was thinking about it and was waiting for this kind of feedback before exploring the option ;)

If we don't pin the nginx version, there is a risk of error as nginx may complain that the module is compiled with different options than the main binary.

content-cache/charmcraft.yaml
```diff
@@ -11,6 +11,48 @@ platforms:
...
+      # Get LUA sources
+      git clone https://github.com/openresty/lua-nginx-module.git --depth 1
```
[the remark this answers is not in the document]
seb4stien: Good point. I'll do if we stick to this approach.

content-cache/charmcraft.yaml
```diff
@@ -11,6 +11,48 @@ platforms:
...
+      # Get upstreal module (the one we are interested in)
+      git clone https://github.com/openresty/lua-upstream-nginx-module.git --depth 1
```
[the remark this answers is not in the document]
seb4stien: I don't think so, but I was a bit confused by the charmcraft build stages and wanted to get early feedback before optimizing this part.
I will have a look.

content-cache/src/nginx_manager.py
```diff
@@ -214,13 +302,26 @@ def _create_status_page_config() -> None:
...
+                NGINX_BACKENDS_STATUS_URL_PATH,
+                nginx.Key("allow", "127.0.0.1"),
+                nginx.Key("deny", "all"),
+                nginx.Key("default_type", "text/plain"),
+                NginxLuaSection(
+                    "content_by_lua_block",
+                    """local hc = require "healthcheck"
+                ngx.say("Nginx Worker PID: ", ngx.worker.pid())
+                ngx.print(hc.status_page())
+                """,
```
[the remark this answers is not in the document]
seb4stien: I will add comment. This is a specific status page that shows the status of the backends as seen by the worker.

content-cache/src/nginx_manager.py
```diff
@@ -299,6 +407,40 @@ def _get_upstream_config_keys(config: LocationConfig) -> tuple[nginx.Key, ...]:
...
+            upstream = "{upstream}",
+            type = "{config.protocol.value}",
+
+            http_req = "GET {config.healthcheck_path} HTTP/1.0\r\nHost: {config.hostname}\r\n\r\n",
+
+            port = {433 if config.protocol.value == "https" else 80},
+            interval = {config.healthcheck_interval},
+            timeout = 1000,
+            fall = 3,
+            rise = 2,
```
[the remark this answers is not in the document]
seb4stien: If yes, I would bring this a different PR ;)

content-cache/src/state.py
yhaliaw: Ok, that seems to be a good value then.
Thanks for the responses.

content-cache/tests/integration/test_healthchecks.py
jdkandersson: Oh I misread this is not commented out code, all good

content-cache/src/nginx_manager.py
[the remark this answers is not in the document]
jdkandersson: I haven't seen POST/ HEAD health checks, I'm not sure it is common practice? That seems to be going more in the direction of testing functionality compared to just checking that the service is up and running. Is it required for any of the initial users?

content-cache/src/nginx_manager.py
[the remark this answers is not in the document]
jdkandersson: Hmm, so what is Lua? I guess that isn't clear to me without knowing the details here

content-cache/tests/integration/test_healthchecks.py
```diff
@@ -0,0 +1,192 @@
...
+    config[PROXY_CACHE_VALID_CONFIG_NAME] = '["200 10s"]'
+    await cache_tester.setup_config(config)
+    await cache_tester.integrate_config()
+    await model.wait_for_idle([app.name, config_app.name], status="active", timeout=10 * 60)
+
+    response = await cache_tester.query_cache(path="/", hostname=hostname, protocol="http")
+    assert response.status_code == 200
+    assert http_ok_message in response.content.decode("utf-8")
+
+    # Sample status page
```
gregory-schiano, line 86: `Expected status page` we'll check for the "IP:PORT UP"

content-cache-backends-config/src/state.py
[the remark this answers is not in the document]
seb4stien: As discussed during the project meeting, we keep a consistent approach with the rest of the file in this PR and may (or may not) switch approach for all configuration options in a later PR.

content-cache/src/nginx_manager.py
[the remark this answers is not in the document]
seb4stien: Lua is the programming language used to enhance nginx: https://www.lua.org/

Regarding the class, I have update the docstring and used a dataclass to avoid the warning from pylint: 6c070adac1bfc3bf20135d22f57eaf42df6d39c6

content-cache/src/nginx_manager.py
seb4stien: As the code is included in a section in nginx, it actually looks better with the current indentation. So I'd keep it like this.

content-cache/src/nginx_manager.py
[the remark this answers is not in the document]
seb4stien: Done in 6c070adac1bfc3bf20135d22f57eaf42df6d39c6

content-cache/src/state.py
seb4stien: Same answer ;)
We leave it as it is for now.

content-cache/src/nginx_manager.py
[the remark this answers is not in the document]
seb4stien: @gregory-schiano do we have contacts in IS for this specific charm to confirm that GET is fine for them?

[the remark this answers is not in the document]
seb4stien: Forget about it. I add it to the agenda for the next meeting with Haw.

content-cache/charmcraft.yaml
```diff
@@ -13,44 +13,37 @@ platforms:
...
+      curl https://raw.githubusercontent.com/openresty/lua-nginx-module/f44c18859050be5af314bae8f05660478af5e31d/src/api/ngx_http_lua_api.h -O --output-dir debian/
+      dpkg-buildpackage -us -uc
+      cp ../libnginx-mod-http-upstream_0.1_amd64.deb /root/stage
     prime:
-      - ngx_http_lua_upstream_module.so
+      - libnginx-mod-http-upstream_0.1_amd64.deb
   nginx-upstream-healthcheck:
     plugin: dump
     source: https://github.com/openresty/lua-resty-upstream-healthcheck.git
+      # source-commit: 7924b2abbcc3269423611e6d7aa66be28f4076f0
```
jdkandersson, line 44: Should this be commented out?

seb4stien: No indeed. This is the current version so it doesn't change the content of the charm, but I will re-run the build & tests to ensure I did not comment it for some issue with charmcraft.

seb4stien: @jdkandersson tests are ok if you can have a look again ;)

content-cache-backends-config/docs/changelog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 2024-12-16
+
+### **Added**
+
+- Support for healthcheck-path and healthcheck-interval.
```
yhaliaw, line 7: Maybe mention these are juju configurations.

content-cache/charmcraft.yaml
```diff
@@ -11,6 +11,41 @@ platforms:
...
+      # Fetch the module to compile
+      repo=lua-upstream-nginx-module
+      rm -Rf $repo
+      git clone https://github.com/openresty/$repo
+      cd $repo
+      git checkout 07b11be7a2e6e1b75209745915a84d281737d4d7
+
+      # Build package
+      cp -R ../debian .
+      curl https://raw.githubusercontent.com/openresty/lua-nginx-module/f44c18859050be5af314bae8f05660478af5e31d/src/api/ngx_http_lua_api.h -O --output-dir debian/
```
yhaliaw, line 36: Line 32 and 36, means we are pinning version right?

I think we need to think about when to update the versions. Since the renovate bot and security check would likely not scan/flag this for upgrades.

No changes needed, just some thought on this.

content-cache/docs/changelog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 2024-12-16
+
+### **Added**
+
+- Support for active healthchecks using LUA scripts. Confofigured through healthcheck-interval and healthcheck-path.
```
yhaliaw, line 7: I think there is a typo `Confofigured`.

content-cache/lua-upstream-nginx-module-packaging/debian/changelog
yhaliaw, line 1: How does the files under `lua-upstream-nginx-module-packaging/debian` work?
Would it be good to have a README, since they are not standard charm src files.

content-cache/src/nginx_manager.py
```diff
@@ -58,6 +66,25 @@
 # This should be tested with integration tests.
 
 
+@dataclass
+class NginxLuaSection:
+    """Simple class to be used by python-nginx to insert Lua code in Nginx conf.
+
+    Attrs:
+        name: the name of the Lua code section to generate (for instance: "content_by_lua_block")
```
yhaliaw, line 74: I think the team generally use full sentence for the docstrings, i.e., Cap the first letter and add a period at the end.

content-cache/src/nginx_manager.py
```diff
@@ -172,17 +225,19 @@ def _load_config() -> None:  # pragma: no cover
     if _systemctl_status_check():
         logger.info("Loading nginx configuration files")
         # This is reload the configuration files without interrupting service.
-        execute_command(["sudo", "nginx", "-s", "reload"])
+        execute_command(["sudo", NGINX_BIN, "-s", "reload"])
```
yhaliaw, line 228: Good catch. Should not have hard coded these values.

content-cache/tests/integration/conftest.py
```diff
@@ -104,8 +104,11 @@ async def deploy_applications_fixture(
                 cert_app_name: model.applications[cert_app_name],
                 metric_app_name: model.applications[metric_app_name],
             }
-        except KeyError:
-            raise RuntimeError("At least one app is missing, you cannot use --no-deploy.")
+        except KeyError as e:
```
yhaliaw, line 107: I think it is team's policy to avoid single character variable names.
So changing `e` to `err` or `exc` might be better.

yhaliaw (commented):
I left some questions and suggestions.
Good work overall.

content-cache-backends-config/docs/changelog.md
seb4stien: Fixed in 2cb65e0b15a811fd42575839098953e78e2d1296

content-cache/charmcraft.yaml
seb4stien: Indeed. As the project is not publishing realeases nor having a clear tagging plan, we thought with Greg it was better to stick to a clear version to not introduce potential (security) issues using latest.

content-cache/docs/changelog.md
seb4stien: Fixed in fd24ec3d7a8d66a860b21810fe8aac018415c9d8

content-cache/src/nginx_manager.py
seb4stien: Thanks. I fixed it in cdf94c57a53954759b47a0f851b893964c36bf05.

content-cache/tests/integration/conftest.py
seb4stien: Fixed in 0a8c9a9ff1805b9f994de0746536685b017e7973. Thanks.

content-cache/src/nginx_manager.py
seb4stien: It's less useful now that we decided to stick to nginx and not switch to openresty. But it was convenient during the tests and if we plan to switch in the future ;)

content-cache/lua-upstream-nginx-module-packaging/debian/changelog
seb4stien: Added some doc in 982c84d6bbc4d7ae89889742bc7e2853289fad72.
Let me know if this is clear enough.

jdkandersson (approved):

yhaliaw (approved):

PR #16, opened 2024-12-11
Bug Fix: Nginx configuration conflict with multiple virtual hosts

yhaliaw:
### Overview

Change how the nginx configuration is setup for proxy cache and log format.
These was causing conflicts with multiple virtual hosts.
The integration test is updated to include test with integration to multiple content-cache-backends-config charm to create multiple virtual hosts in the same content-cache charm.

### Rationale

Bug fix to allow multiple virtual hosts in the same content-cache charm.

### Module Changes

The `nginx_manager.py` is updated with the fixed configuration management. 

### Checklist

- [x] The documentation for charmhub is updated
- [x] The docs/changelog.md is updated with user-relevant changes.

### DIFF
--- a/content-cache/src/nginx_manager.py
+++ b/content-cache/src/nginx_manager.py
@@ -27,6 +27,7 @@ from utilities import execute_command
 logger = logging.getLogger(__name__)
 
 NGINX_CERTIFICATES_PATH = Path("/etc/nginx/certs")
+NGINX_CONF_PATH = Path("/etc/nginx/conf.d")
 NGINX_SITES_ENABLED_PATH = Path("/etc/nginx/sites-enabled")
 NGINX_SITES_AVAILABLE_PATH = Path("/etc/nginx/sites-available")
 NGINX_LOG_PATH = Path("/var/log/nginx")
@@ -138,6 +139,7 @@ def update_and_load_config(
     _reset_nginx_files()
 
     try:
+        _create_http_config()
         _create_status_page_config()
     except NginxFileError:
         logger.info("Stop updating configuration file due to file write issues")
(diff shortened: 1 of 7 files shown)


content-cache/src/nginx_manager.py
```diff
@@ -27,6 +27,7 @@
 logger = logging.getLogger(__name__)
 
 NGINX_CERTIFICATES_PATH = Path("/etc/nginx/certs")
+NGINX_CONF_PATH = Path("/etc/nginx/conf.d")
```
seb4stien, line 30: I'd suggest NGINX_CONFD_PATH or NGINX_CONF_DIRECTORY_PATH as NGINX_CONF_PATH may be used later for /etc/nginx/nginx.conf

content-cache/src/nginx_manager.py
```diff
@@ -217,7 +248,7 @@ def _create_status_page_config() -> None:
             )
         )
     )
-    _create_and_enable_config("nginx_status", nginx_config)
+    _store_and_enable_site_config("nginx_status", nginx_config)
 
 
 def _create_server_config(
```
seb4stien, line 254: As "Server" is an nginx keyword refering to a different section than the one we are generating here. I suggest to be more explicit by renaming the function create_virtualhost_config (or create_site_config). It will make the scope clearer.

content-cache/src/nginx_manager.py
yhaliaw: Updated.
Thanks for the feedback.

content-cache/src/nginx_manager.py
yhaliaw: Updated.

jdkandersson (approved):

alithethird (approved):

PR #32, opened 2025-01-13
Introduce new configuration variables for healthchecks

seb4stien:
Applicable spec: ISD183

### Overview

The current active healthchecks only support 200 as a valid status code. It also only support valid ssl certificate for backends.
It's a limitation as for internal deployments users may have self signed certificates.

This PR proposed to introduce the following configuration options to have more flexibility:
- healthchecks-valid-status: to support more statuses than 200 (e.g. 301)
- healthchecks-ssl-verify: to support self-signed certificates.

### Rationale

With these two additional parameters, SRE should be able to handle all the specific situations they may encounter with the backends.

### Juju Events Changes

None

### Module Changes

None

### Library Changes

None

### Checklist

- [x] The documentation for charmhub is updated
- [x] The docs/changelog.md is updated with user-relevant changes.


content-cache/src/nginx_manager.py
```diff
@@ -477,6 +477,7 @@ def _get_upstream_healthchecks_worker(upstream: str, config: LocationConfig) ->
     Returns:
         A string with the lua script for the healthcheck workers.
     """
+    valid_status_str = ",".join([str(status) for status in config.healthcheck_valid_status])
```
jdkandersson, line 480: `",".join(map(str, config.healthcheck_valid_status))`
`",".join((str(status) for status in config.healthcheck_valid_status))`

content-cache/src/state.py
```diff
@@ -173,6 +179,7 @@ def validate_proxy_cache_valid(cls, value: tuple[str, ...]) -> tuple[str, ...]:
         return value
 
     @classmethod
+    # pylint: disable=too-many-locals
```
jdkandersson, line 182: Similar to earlier comment around disabling linting rules

content-cache/src/state.py
```diff
@@ -191,31 +198,19 @@ def from_integration_data(cls, data: ops.RelationDataContent) -> "LocationConfig
...
-        try:
-            backends = json.loads(backends_str)
-        except json.decoder.JSONDecodeError as err:
-            raise ConfigurationError("Unable to parse backends as json") from err
-
-        if not isinstance(backends, list):
-            raise ConfigurationError("Unable to convert backends to list")
-        if not backends:
-            raise ConfigurationError("Empty backends found")
+        proxy_cache_valid = _load_json_list_field(
```
jdkandersson, line 207: `_parse_list`

content-cache/tests/integration/test_healthchecks.py
```diff
@@ -192,3 +194,78 @@ async def test_healthchecks_all_unhealthy(
...
+    config[PROXY_CACHE_VALID_CONFIG_NAME] = '["200 10s"]'
+    await cache_tester.setup_config(config)
+    await cache_tester.integrate_config()
+    await model.wait_for_idle([app.name, config_app.name], status="active", timeout=10 * 60)
+
+    await asyncio.sleep(5 * HEALTHCHECK_INTERVAL / 1000)
+
+    response = await cache_tester.query_cache(path="/", hostname=hostname, protocol="http")
+    assert response.status_code == 200
+    assert http_ok_message in response.content.decode("utf-8")
```
jdkandersson, line 271: Could this be parametrized into 1 test?

content-cache/tests/integration/test_healthchecks.py
jdkandersson, line 1: Could we test the verify SSL option please?

content-cache/src/state.py
seb4stien: Changed in 15bb1158f452d2ae5242860c7268cdafdba73084

content-cache/tests/integration/test_healthchecks.py
seb4stien: Added in 257b24f97d73d9c7a2ded25556465d2d63f15709.

content-cache-backends-config/src/state.py
```diff
@@ -255,6 +306,12 @@ def to_integration_data(self) -> dict[str, str]:
                 raise ConfigurationError(
                     "Unable to convert configuration to integration data format"
                 ) from err
+
+        for key, value in data[HEALTHCHECK_CONFIG_KEY].items():
+            data["healthcheck_" + key] = str(value).lower()
```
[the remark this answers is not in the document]
seb4stien: I had a look:
- We cannot use a field_serializer at it serializes the value of a field only.
- We can use a model_serializer to do whatever we want, but I'm not convinced that's better than the current approach.

I feel that if we decide to use submodels in the charm, we should stick to it within the charm and not alter default pydantic behaviors.
The fact that we need to flatten the configuration to expose it in the relation is specific to this use case, so it makes sense to me to have a dedicated function as it is today.
However, I can generalize the flattening behavior and remove the constant by just expanding all dicts as it's most likely what we expect.
WDYT?

content-cache-backends-config/tests/unit/test_state.py
```diff
@@ -457,7 +459,7 @@ def test_configuration_to_json_dumps_error(monkeypatch):
     ],
     ids=["empty", "bad_character"],
 )
-def test_invalid_healthcheck_path(bad_value, error_msg):
+def test_invalid_heathcheck_path(bad_value, error_msg):
```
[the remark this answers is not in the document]
seb4stien: Fixed in 947929fa2284375c76601ebc83b08943084e14bb

content-cache/src/state.py
seb4stien: The code has been refactored.

content-cache-backends-config/src/state.py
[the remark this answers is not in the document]
seb4stien: Proposal in 96f8d4b4c9f8e8c84844fed32c32d61c7178a949

content-cache/src/state.py
```diff
@@ -100,6 +102,63 @@ def _validate_path_value(value: str) -> str:
...
+            The object.
+        """
+        interval = int(data.get(HEALTHCHECK_INTERVAL_FIELD_NAME, "-1").strip())
+        path = data.get(HEALTHCHECK_PATH_FIELD_NAME, "").strip()
+        ssl_verify = data.get(HEALTHCHECK_SSL_VERIFY_FIELD_NAME, "").strip()
+        valid_status_str = data.get(HEALTHCHECK_VALID_STATUS_FIELD_NAME, "").strip()
+
+        valid_status = _parse_list(
+            HEALTHCHECK_VALID_STATUS_FIELD_NAME, valid_status_str, raise_if_empty=True
+        )
```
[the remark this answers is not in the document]
seb4stien: Same as above. Not sure if we should change the model_validate_json or use pydantic validators as what we are doing is not related to json but specific to the relations data which is a dict[str:str].

gregory-schiano (approved):
LGTM

alithethird (approved):
Just found 2 typos in docstring. 
LGTM :rocket:

PR #33, opened 2025-01-23
[WIP] Use submodules to track external dependencies

seb4stien:
Explore the possibility to use submodules and renovate to be warned about dependencies updates.


jdkandersson (commented):
Looks fine as long as it works 🙂

PR #37, opened 2025-01-27
Open ports

seb4stien:
Open ports assuming we open port 80 per default and 443 only if we could process certificates successfully.

Port 80 remains open even when 443 is active to redirect clients (as defined in the spec).

### DIFF
--- a/content-cache/src/charm.py
+++ b/content-cache/src/charm.py
@@ -147,6 +147,7 @@ class ContentCacheCharm(ops.CharmBase):
                 self.certificates,
             )
             logger.info("Found all certificate requested")
+            self.unit.open_port("tcp", 443)
         except TLSCertificateIntegrationNotExistError:
             logger.info("Skipping TLS certificates as tls-certificate integration not found")
         except TLSCertificateFileError:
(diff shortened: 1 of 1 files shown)


alithethird (approved):

content-cache/src/charm.py
```diff
@@ -147,6 +147,7 @@ def _load_nginx_config(self) -> None:
                 self.certificates,
             )
             logger.info("Found all certificate requested")
+            self.unit.open_port("tcp", 443)
```
gregory-schiano, line 150: I'm wondering if we shouldn't open the port 443 irrespective of whether the certificates are available or not. Because this is something dynamic, but we never "close" the port. So we might endup in situation where the port is open but not using it, and cases where the port is not opened and we're not using it (let's say I relate to TLS, enable a website, then remove the TLS relation).
So we should probably bind the port to whether there's a TLS certificate relation (add it when the relation is there, and close is if there's no relation). Otherwise easier, always open the port.

seb4stien: I went for the simpler :)
Nowadays https will most likely always be used.

gregory-schiano (approved):

PR #43, opened 2025-01-28
Configure renovate for monorepo

seb4stien:
Renovate configuration was previously in each repo. Now that we have merge, we need a common configuration.
I'm taking the opportunity to introduce git submodules tracking as we will use it to monitor external dependencies.


amandahla:
I didn't follow: why add renovate configuration for rockcraft if the charm does not have rocks?

seb4stien:
Good point. I just migrated the configuration from one of the repo, but I guess it was a copy-paste as we indeed have not rock there.
I will just close the PR for now as my tests on submodules in parallel are not conclusive.

PR #44, opened 2025-01-28
Use submodules to track dependencies

seb4stien:
No changes on the charm itself, we are just using git submodules to track potential changes on dependencies.


seb4stien:
Submodules cannot be used as we are not fetching them during the checkout of the repo. And we cannot do it during the charmcraft build as we are not in a git repository anymore.

PR #67, opened 2025-08-08
Reference the old documentation page

seb4stien:
It seems that we cannot change the discourse link once a charm has been published.

In this PR I update the charm documentation to refer to the existing page, and in parallel I'm switching the content of the odl page and the new page on discourse. The new page will reference the old one.


dimaqq (approved):
Perhaps that's not very elegant, "although practicality beats purity" :)

arturo-seijas (approved):

PR #80, opened 2025-09-17
Doc/contributing

seb4stien:
@erinecon I came up with an hybrid approach for CONTRIBUTING.md:
- everything common is at the top-level
- everything specific remains in the charms

For the how-to contribute, they stay in the charm as are published separately on discourse.

Let me know what you think about it.


CONTRIBUTING.md
```diff
@@ -0,0 +1,110 @@
...
+  [Contributing Guide](https://github.com/canonical/is-charms-contributing-guide).
+
+## Code of conduct
+
+When contributing, you must abide by the
+[Ubuntu Code of Conduct](https://ubuntu.com/community/ethos/code-of-conduct).
+
+## Releases and versions
+
+This project uses [semantic versioning](https://semver.org/).
```
erinecon, line 30: I don't think we'll use semantic versioning for this project (this text has been removed in the charm template repo now). So I think you can remove this line

erinecon (approved):
Thanks -- I think the hybrid approach works well for this monorepo setup. I have one (non-blocking) nitpick but otherwise looks great!

PR #125, opened 2026-06-17
docs: add explanation page for charm design and architecture (ISD-5927)

florentianayuwono:
#### What this PR does

Adds a new Explanation section (Diátaxis) to the documentation with a single page at `docs/explanation/charm-design.md` covering how the charm works under the hood:

- Static-only caching assumption and nginx default cache key behavior (`$scheme$proxy_host$request_uri`)
- Cache storage model: disk for response bodies, shared memory keys zone (10 MB per hostname) for metadata, and LRU eviction behavior
- Backend health checks via `lua-resty-upstream-healthcheck`, failover thresholds (`fall=3`, `rise=2`), and 502 behavior when all backends are down
- TLS termination model and Maintenance status behavior when certificates are expected but not yet issued
- Multi-host isolation: each `content-cache-backends-config` relation gets its own cache directory, keys zone, upstream block, and log files

Also adds `docs/explanation/index.md` and registers the new section in the main `docs/index.md` toctree.

#### Why we need it

Operators need to understand the design opinions behind the charm to predict caching behavior and avoid common issues — for example, knowing that session-cookie-based content will not be cached correctly, or that the charm will not fall back to HTTP if a TLS certificate is delayed.

Tracked in ISD-5927.

#### Test plan

Documentation-only change. No code changes. Verified all nginx-backed claims against the nginx module documentation and all charm behavior claims against the source code (`nginx_manager.py`, `charm.py`, `certificates.py`, `state.py`).

#### Review focus

- Accuracy of nginx behavior descriptions (cache key, keys zone capacity, LRU eviction)
- Accuracy of TLS fallback behavior description
- All nginx doc links point to the correct directive anchors

#### Checklist

- [ ] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)


docs/explanation/index.md
```diff
@@ -0,0 +1,17 @@
...
+(explanation_index)=
+
+# Explanation
+
+These pages explain the design decisions behind the Content Cache Operators and how they affect
+real-world usage.
+
+```{toctree}
+:maxdepth: 1
+Charm design and how it works <charm-design>
```
erinecon:
```suggestion
Charm design <charm-design>
```
Nit, for a shorter name that fits in one line in the left-hand navigation table

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,198 @@
...
+    "description lang=en": "Explanation of the Content Cache charm's design decisions and how they affect caching behavior, memory usage, failover, and TLS."
+---
+
+(explanation_charm_design)=
+
+# Charm design and how it works under the hood
+
+The Content Cache charm makes deliberate, opinionated design decisions to optimize for
+simplicity and static content caching. This page explains those decisions and how they affect
+behavior in real deployments, so that you can predict outcomes and avoid unexpected issues.
```
erinecon:
```suggestion
simplicity and static content caching. Understanding those decisions and how they affect
behavior in real deployments can help you predict outcomes and avoid unexpected issues.
```
Nit, trying to remove the meta-documentation in this sentence (meaning, text that explains the documentation itself)

docs/explanation/charm-design.md
erinecon, line 1: Should it be "nginx" or "NGINX"? Based on https://nginx.org/ it seems like it should be uncapitalized, but based on https://github.com/nginx/nginx it seems like it should be in all-caps

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Static-only caching assumption
+
+The charm is built exclusively for caching static, non-personalized content: image
+assets, CSS files, HTML pages that look the same regardless of who requests them.
```
erinecon:
```suggestion
The charm is built exclusively for caching static, non-personalized content such as image
assets, CSS files, and HTML pages. This content should look the same regardless of who requests it.
```
nits

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,198 @@
...
+nginx identifies a cacheable response by its cache key. The charm does not set a
+[`proxy_cache_key`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_key)
+directive, so nginx uses its default: `$scheme$proxy_host$request_uri`.
+
+`$request_uri` includes the path and any query string. Implication:
+
+- `GET /page?lang=en` and `GET /page?lang=fr` produce different cache keys and are stored
+  as separate cache entries, and therefore query-parameter-based variation works correctly.
+- However, the key does not consider request headers such as `Cookie`, `Authorization`, or
+  `X-User-ID`.
```
erinecon:
```suggestion
`$request_uri` includes the path and any query string. For example,
`GET /page?lang=en` and `GET /page?lang=fr` produce different cache keys and are stored
as separate cache entries, and therefore query-parameter-based variation works correctly.
The key does not consider request headers such as `Cookie`, `Authorization`, or
`X-User-ID`.
```
Nits, I'm trying to reduce the amount of bulleted lists in the document, and I think this is a good place to move the list into the paragraph.

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,198 @@
...
+
+This means that two requests with the same URL but different session cookies will share a
+single cache entry. The first response is cached and served to every subsequent requester of
+that URL, regardless of their session or identity. For personalized or session-dependent
+content, this produces incorrect results.
```
erinecon:
```suggestion
content, this behavior produces incorrect results.
```
Trying to avoid a lone "this", so that the sentence is explicit about what is producing incorrect results. If it's not the "behavior", maybe it's the design?

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,198 @@
...
+}
+```
+
+The [`proxy_cache`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache)
+directive (set at the server block level) ties this location to its hostname's
+dedicated cache zone.
+
+## Cache storage: disk and RAM
+
+nginx uses a two-tier storage model for caching:
```
erinecon:
```suggestion
## Cache storage

nginx uses a two-tier storage model for caching: disk and RAM.
```
nit

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,198 @@
...
+```nginx
+proxy_cache_path /data/nginx/cache/example.com
+    use_temp_path=off
+    levels=1:2
+    keys_zone=example.com:10m;
+```
+
+The 10 MB limit is fixed in the charm and cannot be changed via configuration. For most static
+content deployments this is sufficient: per the nginx docs, one megabyte zone can store about
+8,000 keys, meaning 10 MB supports approximately 80,000 cached entries.
```
erinecon:
```suggestion
content deployments this is sufficient: per the nginx docs,
10 MB supports approximately 80,000 cached entries.
```
nit, for conciseness

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,198 @@
...
+
+### What happens when the keys zone fills up
+
+When the keys zone is full, nginx applies
+[LRU (Least Recently Used) eviction](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_path):
+the metadata entry for the least recently accessed cache item is removed from the shared memory zone.
+
+### Disk expiry and cache lifetime
+
+Disk entries are expired according to
```
erinecon:
```suggestion
Disk entries expire according to
```

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,198 @@
...
+[`proxy_cache_valid`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_valid),
+which maps HTTP response codes to TTLs. For example:
+
+```
+proxy-cache-valid: '["200 302 1h", "404 1m"]'
+```
+
+This caches 200 and 302 responses for one hour, and 404 responses for one minute. Responses
```
erinecon:
```suggestion
This directive caches 200 and 302 responses for one hour, and 404 responses for one minute. Responses
```
Nit, for clarity. If it's not the directive that caches the responses, please let me know!

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,198 @@
...
+### All backends unavailable
+
+If all backends for a hostname are simultaneously marked down (either by the health checker or
+by `fail-timeout`), nginx returns 502 Bad Gateway to the client. Cached content for the
+affected paths may still be served if the entries are still valid according to `proxy-cache-valid`.
+nginx does not serve stale content beyond its TTL by default.
+
+## TLS termination
+
+TLS is terminated at the nginx instance managed by this charm. Backends are always addressed
```
erinecon: Is the TLS terminated *at* the nginx instance, or *by* the nginx instance? Asked in another way: Is the nginx instance the thing doing the termination, or does the TLS remain alive for other instances and is terminated only for the nginx instance?

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,198 @@
...
+directly by IP address over the protocol specified by the `protocol` configuration option
+(`http` or `https`).
+
+TLS certificates are obtained via the Juju `certificates` integration
```
erinecon:
```suggestion
TLS certificates are obtained via the Juju `certificates` relation
```

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,198 @@
...
+(using the `tls-certificates` interface). When a `content-cache-backends-config` relation
+provides a hostname, the charm requests a certificate for that hostname.
+
+### Behavior when certificates are not yet available
+
+If the `certificates` integration exists but the certificate for a hostname has not yet been
```
erinecon:
```suggestion
If the `certificates` relation exists but the certificate for a hostname has not yet been
```

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,198 @@
...
+issued, the charm enters `Maintenance` status and does not load any nginx configuration
+until all required certificates are available. It will not fall back to serving the hostname
+over plain HTTP.
+
+This is an intentional security decision, as a charm that has been told to expect TLS should not
+silently serve unencrypted traffic because a certificate is delayed.
+
+If no `certificates` integration is present, the charm does not add a `listen 443 ssl` directive
```
erinecon:
```suggestion
If no `certificates` relation is present, the charm does not add a `listen 443 ssl` directive
```

erinecon (commented):
Thank you so much for creating this explanation document! 🎉

docs/explanation/charm-design.md
florentianayuwono: im unsure on that aspect, but NGINX made it sound like it's a proper name on its own. Although it also made it sound like it is an abbreviation for each of the capitalized words, when it's actl just how it's engine x is pronounced?

docs/explanation/charm-design.md
florentianayuwono: behavior is the right term

docs/explanation/charm-design.md
erinecon: Yeah that's a fair point. I think leaving it uncapitalized is okay (the upstream docs leaves the name uncapitalized, anyway), and you were consistent about it, so no need to change anything. Thanks!

docs/explanation/charm-design.md
erinecon: FWIW I think that this sentence should say:
```suggestion
TLS is terminated by the nginx instance managed by this charm. Backends are always addressed
```
Because my understanding is that the instance is the one doing the termination. But please correct me if I'm mistaken!

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,200 @@
...
+### Multi-host isolation
+
+Each hostname configured via a `content-cache-backends-config` relation gets:
+
+- Its own cache directory (`/data/nginx/cache/<hostname>/`)
+- Its own RAM keys zone (`keys_zone=<hostname>:10m`)
+- Its own upstream block and log files
+
+There is no cross-hostname competition for RAM. Each hostname has its own `keys_zone`
+allocation, so one hostname's cache metadata cannot evict another's. Disk capacity, however,
```
florentianayuwono:
```suggestion
allocation, so one hostname's cache metadata cannot evict other cache. Disk capacity, however,
```

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,200 @@
...
+behavior in real deployments can help you predict outcomes and avoid unexpected issues.
+
+## Static-only caching assumption
+
+The charm is built exclusively for caching static, non-personalized content such as image
+assets, CSS files, and HTML pages. This content should look the same regardless of who requests it.
+
+nginx identifies a cacheable response by its cache key. The charm does not set a
+[`proxy_cache_key`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_key)
+directive, so nginx uses its default: `$scheme$proxy_host$request_uri`.
```
erinecon:
```suggestion
directive, so nginx uses the default `$scheme$proxy_host$request_uri`.
```
nit

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,200 @@
...
+    use_temp_path=off
+    levels=1:2
+    keys_zone=example.com:10m;
+```
+
+The 10 MB limit is fixed in the charm and cannot be changed via configuration. For most static
+content deployments this is sufficient: per the nginx docs,
+10 MB supports approximately 80,000 cached entries.
+
+### What happens when the keys zone fills up
```
erinecon:
```suggestion
### When the keys zone fills up
```
nit, for conciseness

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,200 @@
...
+
+The charm uses the [lua-resty-upstream-healthcheck](https://github.com/openresty/lua-resty-upstream-healthcheck)
+module to actively monitor backend health. A Lua worker runs inside each nginx worker process
+and periodically probes each backend in the background.
+
+The health check parameters are configured per relation:
+
+| Parameter | Description | Default |
+|---|---|---|
+| `healthcheck-interval` | Time between checks (ms) | 10000 |
```
erinecon:
```suggestion
| `healthcheck-interval` | Time between checks (ms) | `10000` |
```
nit, for consistency with the other entries in the table

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,200 @@
...
+| `healthcheck-valid-status` | HTTP codes considered healthy | `200` |
+| `healthcheck-ssl-verify` | Verify SSL cert on HTTPS checks (set to `false` to skip verification) | `true` |
+
+The checker uses fall/rise thresholds to avoid flapping:
+
+- A backend is marked down after 3 consecutive failures (`fall=3`)
+- A backend is marked up again after 2 consecutive successes (`rise=2`)
+
+The generated Lua block for a single backend looks like (using non-default values for
+`healthcheck-path` and `protocol` as an example):
```
erinecon:
```suggestion
The following example shows the generated Lua block for a single backend
using non-default values for `healthcheck-path` and `protocol`:
```

erinecon (approved):
Nits only on my end -- Thanks so much!

Once Maskim gives a :+1:, we should be good to merge!

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,200 @@
...
+- Pages that vary by logged-in user (e.g. dashboards, account pages)
+- API responses that differ based on cookies or auth tokens (same URL, different users)
+- Any content where the correct response depends on the identity of the user making the request
+
+It is well-suited for:
+
+- Public marketing pages and blog posts
+- Static asset files (JS, CSS, fonts, images)
+- Documentation sites
+- Content that is identical for every visitor, or that varies only by URL/query parameters
```
gregory-schiano: Public marketing pages, blog posts, documentation, ... are well-suited because they mostly contain static asset files.
So I'd remove the "Static asset files" as it's more a technical description, while the others are business case descriptions.

Maybe
`It is well-suited for static asset files (JS, CSS, fonts, images), found in such uses cases:` (or something around the line)

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,200 @@
...
+
+- Its own cache directory (`/data/nginx/cache/<hostname>/`)
+- Its own RAM keys zone (`keys_zone=<hostname>:10m`)
+- Its own upstream block and log files
+
+There is no cross-hostname competition for RAM. Each hostname has its own `keys_zone`
+allocation, so one hostname's cache metadata cannot evict other cache. Disk capacity, however,
+is shared across all hostnames on the same filesystem. Adding or removing a
+`content-cache-backends-config` relation only affects that hostname's configuration; other
+hostnames continue serving from their own caches uninterrupted.
```
gregory-schiano, line 123: We should add a warning that when disk is full, Nginx will throw errors (TBC), therefore a good attention to the disk size if important (especially in the context of caching big files)

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,200 @@
...
+
```
gregory-schiano, line 124: Maybe investigate when nginx cleans up the disk, is it done immediately when a key is evicted, or when a new entry with the same position is added and add some explanation.

docs/explanation/charm-design.md
beliaev-maksim: should we mention binaries, tars?
I am thinking of stores and other use cases

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,200 @@
...
+The generated nginx location block looks like:
+
+```nginx
+location / {
+    proxy_pass http://<upstream-uuid>/;
+    proxy_set_header Host "example.com";
+    proxy_cache_valid 200 302 1h;
+    proxy_cache_valid 404 1m;
+}
+```
```
beliaev-maksim: for me that comes out of the blue :)
can you please add some context around?

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,200 @@
...
+the metadata entry for the least recently accessed cache item is removed from the shared memory zone.
+
+### Disk expiry and cache lifetime
+
+Disk entries expire according to
+[`proxy_cache_valid`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_valid),
+which maps HTTP response codes to TTLs. For example:
+
+```
+proxy-cache-valid: '["200 302 1h", "404 1m"]'
```
beliaev-maksim, line 105: is that the charm config that users can apply ?

so the response code's TTLs for all the URLs will be shared?

beliaev-maksim: I see it below, but maybe we can reconcile and mention that this data is consumed by the charm via relation

@erinecon

docs/explanation/charm-design.md
beliaev-maksim: @gregory-schiano do we provide observability of the disk space? or it is outside of this charm?

docs/explanation/charm-design.md
```diff
@@ -0,0 +1,200 @@
...
+
+If all backends for a hostname are simultaneously marked down (either by the health checker or
+by `fail-timeout`), nginx returns 502 Bad Gateway to the client. Cached content for the
+affected paths may still be served if the entries are still valid according to `proxy-cache-valid`.
+nginx does not serve stale content beyond its TTL by default.
+
+## TLS termination
+
+TLS is terminated by the nginx instance managed by this charm. Backends are always addressed
+directly by IP address over the protocol specified by the `protocol` configuration option
```
beliaev-maksim, line 184: one config for the entire charm? all hostnames use it then?

context: still thinking of our deployment, one cache vs many caches

docs/explanation/charm-design.md
florentianayuwono: caching behavior is further documented in https://github.com/canonical/content-cache-operator/pull/127 currently there's no observability on the disk space

docs/explanation/charm-design.md
florentianayuwono: yep it will be explained in https://github.com/canonical/content-cache-operator/pull/127

docs/explanation/charm-design.md
florentianayuwono: clarified below that one cert per hostname, each relation trigger a separate request

docs/explanation/charm-design.md
gregory-schiano: Disk observability is part of the system observability, we could eventually check if there is a metrics of current cache size. Adding the disk metrics will require a specific system metric collector start by content cache, that could be in conflict with the system metric collector
We should probably document the metrics collected.

florentianayuwono: the metrics collected hv been documented under an existing enable-cos.md. however regarding disk observability, currently the node_exporter collect the node_filesystem_avail_bytes for all mounted filesystems, but cache disk usage will not be separately visible unless we manually mount a volume at /data/nginx/cache/

PR #127, opened 2026-06-23
docs: add explanation page for caching behavior (ISD-5928)

florentianayuwono:
#### What this PR does

Adds a new Explanation page `docs/explanation/caching-behavior.md` covering how the charm caches static content, with a focused section on large-file (e.g. Ubuntu ISO image) caching considerations.

The page covers:
- Cache miss and hit flow (how nginx fetches, writes to disk, and streams to the client simultaneously)
- RAM vs disk roles (`keys_zone` stores metadata only; file bodies live on disk)
- Cache expiry (TTL via `proxy-cache-valid`) and eviction (nginx `inactive` default of 10 minutes) as two independent mechanisms
- Concurrent first-hit behavior (no `proxy_cache_lock` set)
- Large-file-specific considerations: disk capacity (no `max_size` or `min_free` set, so no LRU eviction), inactive eviction impact, and concurrent bandwidth amplification
- Quick-reference summary table

Also adds `docs/explanation/index.md` and registers the explanation section in `docs/index.md`.

#### Why we need it

Operators caching large binary files need to understand the disk capacity risk, inactive eviction behavior, and concurrent-fetch behavior. Without this page, those behaviors are undocumented and operators may be surprised when the disk fills up or repeated upstream fetches occur for infrequently-accessed large files.

All claims are source-verified against `content-cache/src/nginx_manager.py` and the nginx official documentation.

#### Test plan

- `make spelling`: passes (added `ISOs`, `uncached`, `toctree` to `.custom_wordlist.txt`)
- `make lint-md`: passes

#### Review focus

- **Disk-full behavior**: the docs state nginx fails to write new cache entries (no LRU eviction) because neither `max_size` nor `min_free` is set. This is sourced from nginx docs.
- **Inactive timeout**: stated as 10-minute nginx default, sourced from nginx docs. No charm config option to change it — verified by absence of `inactive=` in `nginx_manager.py` line 402.
- **`use_temp_path=off`**: sourced from `nginx_manager.py` line 402.

#### Checklist

- [x] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)


docs/explanation/caching-behavior.md
```diff
@@ -0,0 +1,116 @@
...
+
+- Provision a dedicated large volume mounted at `/data/nginx/cache/` before deploying for
+  large-file caching.
+- Monitor disk usage and set up alerts before the filesystem fills up.
+
+When the disk fills up, nginx fails to write new cache entries. Existing cached files remain
+accessible, but requests for uncached files will fail if nginx cannot write the response to
+disk during the upstream fetch.
+
+### Inactive eviction impact
```
beliaev-maksim, line 112: is there a way to control this per file?
we indeed have this exact issue when some clouds start the refresh at the same time, so we need to spread the load

docs/explanation/caching-behavior.md
```diff
@@ -0,0 +1,116 @@
...
+
+The charm does not enable
+[`proxy_cache_lock`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_lock).
+When multiple clients simultaneously request the same URL that is not yet cached, nginx
+starts a separate upstream fetch for each request rather than having the first fetch complete
+while the rest wait.
+
+For small files this is usually acceptable — the window of concurrent cache misses is brief
+and the bandwidth consumed is modest. For large files the impact is more significant (see
+[Large-file considerations](#large-file-considerations) below).
```
erinecon: This cross-reference is throwing an error when RTD tries to build the project:
```
/home/docs/checkouts/readthedocs.org/user_builds/canonical-content-cache-charms/checkouts/127/docs/explanation/caching-behavior.md:64: WARNING: 'myst' cross-reference target not found: 'large-file-considerations'
```
I suggest that we remove this cross-reference since it's the next section in the document:

```suggestion
the next section).
```

docs/.custom_wordlist.txt
```diff
@@ -1,2 +1,9 @@
 AI
+cacheable
 classDef
+hostnames
+ISOs
+Lua
+toctree
```
erinecon: Why was `toctree` added to the custom wordlist?

docs/explanation/caching-behavior.md
```diff
@@ -0,0 +1,116 @@
...
+
+## How caching works
+
+### Cache miss: first request for a URL
+
+When nginx receives a request for a URL that is not in the cache, it forwards the request to
+an upstream backend. As the response arrives, nginx simultaneously:
+
+1. Streams the response body to the client.
+2. Writes the response body to disk at `/data/nginx/cache/<hostname>/`.
```
erinecon: Since these actions are simultaneous, I think we should use a bulleted list. A numbered list implies that nginx follows this exact order in the actions

docs/explanation/caching-behavior.md
```diff
@@ -0,0 +1,116 @@
...
+These two mechanisms are independent. The inactive timeout can evict a response before its
+TTL expires, and a long TTL does not prevent eviction if the content is not accessed.
+
+There is currently no charm configuration option to change the `inactive` timeout directly.
+
+## Concurrent first-hit requests
+
+The charm does not enable
+[`proxy_cache_lock`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_lock).
+When multiple clients simultaneously request the same URL that is not yet cached, nginx
```
erinecon:
```suggestion
When multiple clients simultaneously request the same uncached URL, nginx
```
nit, for conciseness

docs/explanation/caching-behavior.md
```diff
@@ -0,0 +1,116 @@
...
+starts a separate upstream fetch for each request rather than having the first fetch complete
+while the rest wait.
+
+For small files this is usually acceptable — the window of concurrent cache misses is brief
```
erinecon:
```suggestion
For small files this behavior is usually acceptable — the window of concurrent cache misses is brief
```
nit, to remove a lone "this"

docs/explanation/caching-behavior.md
```diff
@@ -0,0 +1,116 @@
...
+| Behavior | Detail |
+|---|---|
+| Cache miss | nginx fetches from upstream and writes to disk simultaneously |
+| Cache hit | Served from disk; no upstream contact |
+| Write path on first fetch | Direct to final cache location (`use_temp_path=off`) |
+| Disk cap | None — `max_size` is not set |
+| Cache TTL per status code | Configurable via `proxy-cache-valid` on `content-cache-backends-config` |
+| Cache eviction (inactive timeout) | nginx default: 10 minutes of no access; not configurable via the charm |
+| Concurrent first-hit requests | Each triggers a separate upstream fetch (`proxy_cache_lock` not set) |
+| Disk full behavior | New cache writes fail; existing cached files remain accessible |
```
erinecon: Nit: This section feels out of place in an Explanation document. Maybe because the section is titled "Quick reference" -- although the table itself is much more oriented toward reference material regardless of the section title. I suggest we either remove this section entirely or rename to something like `## Summary of behaviors`

erinecon (approved):
Nits only on my end, so giving the PR an approval! Feel free to ping me for another review if the content is substantially updated after incorporating Maksim's comment :)

docs/explanation/caching-behavior.md
florentianayuwono: we can't control this per file, but a workaround is we can change the charm to have cache zone per path (instead of per hostname) and allow the inactive value to be configured so we can have longer inactive window for large file

docs/.custom_wordlist.txt
florentianayuwono: good catch removed!

docs/explanation/caching-behavior.md
florentianayuwono: removed!

docs/explanation/caching-behavior.md
gregory-schiano, line 1: Check if content-cache respects the typical HTTP cache headers (`cache-control`, `pragma`, `expires`, `etag`, ...) and document this

docs/explanation/caching-behavior.md
gregory-schiano: See my other comment, check the cache headers behavior

docs/explanation/caching-behavior.md
gregory-schiano: And if cache headers are respected, how the "inactive" behaves ? Does it apply only for queries without cache headers

florentianayuwono: a section has been added to document this. tldr there r several headers that r rescpected n actl take higher priority, but inactive still applies regardless of these headers

docs/explanation/caching-behavior.md
gregory-schiano: @beliaev-maksim After investigations by Florentiana, the server cache headers are supported by nginx, in order to control what can / can't be cached, but also for how long. These cache headers influence the TTL of the cache, meaning that even if the endpoint is active, Nginx will reply that resource with the cached resource without querying the backend.

These headers are not only meant to control the proxy cache, but also the browser. So when the server tell you to cache some data (e.g. CSS) for 10years, the proxy can't obey by that rule because it's restricted in cache capacity.
Let's take the example of 1 query execute only once, and never again. Nginx will cache that query, and can't keep this data for ever. That's what the inactive is meant for... flushing ressources that are cached but never accessed.

The parameter is 10m by default on nginx, and not configurable for now. This is not meant to control per resource or path, this is meant to control the cache bucket usage. If the backend wants to control per resource, it needs to use the cache headers, but there's no guanratee that this cache will last the requested amount in nginx

The other eviction scenario is when the cache is full, regardless of the backend cache header request, ngninx will evict the least used resource, in order to free space (even if the new ressource that will added, is never asked again, as it can't anticipate future requests)

TLDR;
TTL / Cache-Header: How long do I wait to query back the backend for refreshed information, regardless of the activity
Inactivity: Evicting cache that is not served, regardless of their freshness

PR #129, opened 2026-06-24
docs: add security explanation page (ISD-5931)

florentianayuwono:
#### What this PR does

Adds a new Explanation page `docs/explanation/security.md` covering the security design decisions of the content-cache charm, organised by security boundary.

The page covers:

**Client to charm**
- TLS via the `certificates` relation (HTTPS); without the relation nginx serves HTTP on port 80
- No client authentication — intentional design for publicly accessible static content; operators must place a reverse proxy or WAF in front for access control
- No rate limiting — operators must add this at a component in front of the charm

**Charm to backend**
- `protocol` config defaults to `https`
- `healthcheck-ssl-verify` defaults to `true` but only covers Lua healthcheck pings — `proxy_pass` does not verify backend SSL certificates (`proxy_ssl_verify` defaults to `off` and is not overridden by the charm)

**Internal**
- nginx worker processes run as `www-data` (Ubuntu nginx package default); all cache and certificate files owned by `www-data`
- TLS private keys stored at `/etc/nginx/certs/<hostname>.pem` with permissions `0o644` (readable by local users)
- Status pages (`/nginx_status`, `/nginx_backends_status`) restricted to `127.0.0.1` via `allow`/`deny` directives — hardcoded, not configurable
- `shell=False` on all subprocess calls

**Cached data risks**
- Cache key does not include `Cookie` or `Authorization` headers — personalized content must not be routed through the charm
- No built-in cache purge mechanism

**Best-practices summary table** closing the page.

All claims source-verified against `nginx_manager.py`, `certificates.py`, `config.yaml`, and nginx official documentation.

#### Why we need it

The charm has no existing documentation on security. Operators need to understand what the charm protects, what it does not, and what they must add externally before deploying in production.

#### Test plan

- `make spelling`: passes
- `make lint-md`: passes

#### Review focus

- **`proxy_ssl_verify` gap**: `protocol=https` encrypts backend traffic but does not verify backend certificates. This is sourced from the nginx `proxy_ssl_verify` docs (defaults to `off`) and confirmed by absence of the directive in the charm code.
- **`www-data` claim**: nginx running as `www-data` is from the Ubuntu nginx package default, not set by the charm. The charm creates files owned as `www-data` (sourced from `nginx_manager.py` lines 43, 293–294 and `certificates.py` lines 101, 103).
- **PEM file permissions `0o644`**: sourced from `certificates.py` line 104. This means private keys are readable by any local user — called out explicitly.

#### Checklist

- [x] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)


danielvnguyen (approved):

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+
+The charm provides no client authentication mechanism. Any client that can reach the charm's HTTP(S) listener (port 80 when no `certificates` relation is present, otherwise port 443) can request cached content. This is an intentional design constraint: the charm is built for publicly accessible static content. There are no plans to add a native authentication feature.
+
+### Rate limiting
+
+The charm does not configure nginx rate limiting
+([`limit_req`](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html) or
+[`limit_conn`](https://nginx.org/en/docs/http/ngx_http_limit_conn_module.html)). Operators
+who need protection against abuse or denial-of-service attacks must add rate limiting at a
+component placed in front of the charm, such as a load balancer, reverse proxy, or WAF.
```
erinecon: Nit: Should we define WAF here? Will users already know about WAF, or is there a chance that they won't be familiar?

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Charm to backend
+
+### Backend protocol
+
+The `protocol` configuration option on `content-cache-backends-config` controls whether nginx
+contacts backends over HTTP or HTTPS. It defaults to `https`. Operators should keep this
```
erinecon:
```suggestion
contacts backends over HTTP or HTTPS. The configuration defaults to `https`. Operators should keep this
```
Unless "it" means the charm. Either way, I think we should be more explicit here

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+default unless backends do not support HTTPS.
+
+### Backend SSL certificate verification
+
+The charm contacts backends over two separate code paths: the Lua healthcheck module (periodic
+health pings) and the nginx `proxy_pass` directive (actual proxied requests). These have
+different SSL verification behavior.
+
+**Healthchecks** — the `healthcheck-ssl-verify` option on `content-cache-backends-config`
```
erinecon:
```suggestion
**Healthchecks** — the `healthcheck-ssl-verify` configuration option on `content-cache-backends-config`
```
nit

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+controls whether the Lua healthcheck module verifies the backend SSL certificate during health
+pings. It defaults to `true`. Setting it to `false` disables certificate verification for
```
erinecon:
```suggestion
pings. The configuration defaults to `true`. Setting the configuration to `false` disables certificate verification for
```
Unless "it" means the charm (maybe it makes more sense that it's the charm that has defaults, but you set the configuration?)

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+healthchecks and should only be used in controlled environments — for example, when backends
+use self-signed certificates on a trusted private network.
+
+**Proxied requests** — nginx's
+[`proxy_ssl_verify`](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ssl_verify)
+defaults to `off` and the charm does not override it. This means that even when
+`protocol=https`, nginx encrypts the connection to the backend but does **not** verify the
+SSL certificate of the backend. Traffic to the backend is protected against passive eavesdropping
+but not against a man-in-the-middle attack on that connection. There is no charm configuration
```
erinecon: The inclusive language checker is complaining about "main-in-the-middle". Could we update to something like:
```suggestion
but not against a machine-in-the-middle attack on that connection. There is no charm configuration
```
Otherwise we'll need to place an ignore flag

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+The charm creates and owns all cache and certificate files as `www-data`. The cache directory
+at `/data/nginx/cache/` and the certificate files at `/etc/nginx/certs/` are owned by
+`www-data` (set via `os.chown` in the charm code).
+
+### Status page access
+
+The nginx status page at `/nginx_status` and the backend health status page at
+`/nginx_backends_status` are restricted to `127.0.0.1` only using the nginx
+[`allow`/`deny` directives](https://nginx.org/en/docs/http/ngx_http_access_module.html).
+External clients cannot access these endpoints. This is hardcoded in the generated nginx
```
erinecon:
```suggestion
External clients cannot access these endpoints. This behavior is hard-coded in the generated nginx
```
I think we should be more explicit about what "this" is referring to. I think it's the behavior, but let me know if I misunderstood!

Also I think that we should hyphenate `hardcoded` instead, at least from the two style guides I read ([Salesforce](https://developer.salesforce.com/docs/atlas.en-us.salesforce_pubs_style_guide.meta/salesforce_pubs_style_guide/style_hard_coding.htm), [Microsoft](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/h/hard-code-hard-coded-hard-coding))

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Public-only content
+
+The charm's cache key is based on `$scheme$proxy_host$request_uri` (the nginx default). It
+does not include request headers such as `Cookie` or `Authorization`. Two requests for the
+same URL with different session cookies share a single cache entry: the first response is
+cached and served to every subsequent requester of that URL, regardless of their identity.
+
+Operators must ensure that only public, non-personalized content is routed through the charm.
+Routing personalized or session-dependent content through the charm will cause users to
```
erinecon:
```suggestion
Routing personalized or session-dependent content through the charm means users will
```
nit, no strong opinion

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+receive each other's responses.
+
+### No cache purge
+
+The charm provides no built-in mechanism to purge a cached response on demand — for example,
+if sensitive content was accidentally cached. Cache entries are removed when their
+`proxy-cache-valid` TTL expires or when they are removed by nginx's default `inactive=10m`
+behavior for `proxy_cache_path` (the charm does not override this). Operators who need on-demand purge capability must implement it directly at the nginx level
+outside the charm, or wait for natural expiry.
```
erinecon: Is there a link to upstream docs on how to implement the on-demand purge capability?

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Best practices
+
+| Practice | Recommendation |
+|---|---|
+| Enable TLS | Integrate with a `tls-certificates` provider charm |
```
erinecon:
```suggestion
| TLS | Integrate with a `tls-certificates` provider charm to enable TLS |
```
Nit, to be consistent with the other rows (the first column is object-oriented rather than action-oriented)

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+| Backend protocol | Keep `protocol=https` (the default) |
+| Backend SSL verification | Keep `healthcheck-ssl-verify=true` (the default) |
+| Access control | Place an authenticating reverse proxy or WAF in front if the content is not fully public |
+| Rate limiting | Add rate limiting at a component placed in front of the charm (load balancer, reverse proxy, or WAF) if abuse protection is needed |
+| Cached content | Only route public, non-personalized content through the charm |
+| Machine access | Restrict local user access to the Juju machine — TLS private keys are readable by local users |
```
erinecon:
```suggestion
| Machine access | Restrict local user access to the Juju machine |
```
Nit, to be consistent with the other rows (the other rows didn't provide more context beyond the recommendation). 

I think it's already mentioned that TLS private keys are readable by local users (in the `Transport security` section), so we should be okay to drop it here

docs/.custom_wordlist.txt
```diff
@@ -1,2 +1,11 @@
 AI
+chmod
 classDef
+hardcoded
```
erinecon: I think "hardcoded" should be removed from the custom wordlist (and hyphenated when it appears in the text)

erinecon (commented):
Thank you so much for writing comprehensive security documentation for the content cache charms!

docs/explanation/security.md
florentianayuwono: there are references of how this is done, for example https://github.com/FRiCKLE/ngx_cache_purge and https://www.f5.com/products/nginx/nginx-plus but these r third party/commercial solutions that's not available in the ubuntu nginx package currently

gregory-schiano: I would rephrase, these's no option to purge cache from backend perspective, only waiting for expiration

docs/explanation/index.md
```diff
@@ -0,0 +1,17 @@
+---
+myst:
+  html_meta:
+    "description lang=en": "Explanation of the Content Cache Operators design, architecture, and how they work under the hood."
```
erinecon:
```suggestion
    "description lang=en": "Explanation of the Content Cache Operators design, architecture, and security."
```
nit, think it would be good to include something about security in the metadata

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+  html_meta:
+    "description lang=en": "Explanation of the security design decisions in the Content Cache charm, including TLS, backend authentication, access control, and cached data risks."
+---
+
+(explanation_security)=
+
+# Security
+
+The charm makes deliberate security decisions at three boundaries: the connection from clients
+to nginx, the connection from nginx to backends, and the nginx process itself. This page
```
erinecon: Nit, but the way you describe the boundaries here doesn't totally align with the section titles. I think it's fair to assume that "connection from clients to nginx" corresponds to the "Client to charm" section etc., but that mapping might not be immediately obvious to some users. 

I don't know whether we need to clarify anything, or if it's even accurate to rephrase to something like:
```suggestion
The charm makes deliberate security decisions at three boundaries: the connection from clients
to the charm through nginx, the connection from nginx to backends, and the nginx process itself. This page
```

I guess I'm wondering, do we need to explicitly write out the mapping between nginx and the charm itself? Or are they interchangeable for the purposes of this document?

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+### Backend SSL certificate verification
+
+The charm contacts backends over two separate code paths: the Lua healthcheck module (periodic
+health pings) and the nginx `proxy_pass` directive (actual proxied requests). These have
+different SSL verification behavior.
+
+**Healthchecks** — the `healthcheck-ssl-verify` configuration option on `content-cache-backends-config`
+controls whether the Lua healthcheck module verifies the backend SSL certificate during health
+pings. The configuration defaults to `true`. Setting the configuration to `false` disables certificate verification for
+healthchecks and should only be used in controlled environments — for example, when backends
```
erinecon:
```suggestion
healthchecks and should only be used in controlled environments, for example, when backends
```
nit

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+without a separate exploit.
+
+The charm creates and owns all cache and certificate files as `www-data`. The cache directory
+at `/data/nginx/cache/` and the certificate files at `/etc/nginx/certs/` are owned by
+`www-data` (set via `os.chown` in the charm code).
+
+### Status page access
+
+The nginx status page at `/nginx_status` and the backend health status page at
+`/nginx_backends_status` are restricted to `127.0.0.1` only using the nginx
```
erinecon:
```suggestion
`/nginx_backends_status` are restricted to `127.0.0.1` using the nginx
```
Nit, I don't think "only" adds much value

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+### Subprocess security
+
+The charm uses `shell=False` for all subprocess calls, preventing shell injection attacks from
+malformed configuration values.
+
+## Cached data risks
+
+### Public-only content
+
+The charm's cache key is based on `$scheme$proxy_host$request_uri` (the nginx default). It
```
erinecon:
```suggestion
The charm's cache key is based on the nginx default `$scheme$proxy_host$request_uri`. It
```
nit, I think this flows a little more nicely

docs/explanation/security.md
```diff
@@ -0,0 +1,129 @@
...
+behavior for `proxy_cache_path` (the charm does not override this). Operators who need on-demand purge capability must implement it directly at the nginx level
+outside the charm, or wait for natural expiry.
+
+## Best practices
+
+| Practice | Recommendation |
+|---|---|
+| TLS | Integrate with a `tls-certificates` provider charm to enable TLS |
+| Backend protocol | Keep `protocol=https` (the default) |
+| Backend SSL verification | Keep `healthcheck-ssl-verify=true` (the default) |
```
erinecon:
```suggestion
| Backend protocol | Keep the default `protocol=https` |
| Backend SSL verification | Keep the default `healthcheck-ssl-verify=true` |
```
nit

erinecon (approved):
Nits only, thanks!

docs/explanation/security.md
florentianayuwono: the charm handles the configuration of nginx, so the correct phrase will be client to charm and charm to backend, since the charm controls the nginx

PR #134, opened 2026-07-14
feat: remove hostname/path routing, add per-relation port allocation

florentianayuwono:
## Summary

removes hostname/path-based routing from the `content-cache` and `content-cache-backends-config` charms, replacing it with per-relation unique port allocation on nginx.

### content-cache
- **`state.py`**: Removed `hostname`, `path`, `backends_path` fields from `LocationConfig`; changed `NginxConfig` to a flat `dict[int, LocationConfig]` (relation_id → config); removed unused type aliases (`Hostname`, `Location`, `HostConfig`) and field name constants
- **`nginx_manager.py`**: Each relation now gets its own nginx server block listening on a unique port; replaced path-loop with single `location /`; removed `hostname_to_cert` parameter and TLS cert handling; removed `Host` header from healthcheck Lua script
- **`charm.py`**: Added `StoredState` with `port_map`; added `_get_port_for_relation()` allocating ports from range 8080–8279; removed all TLS certificate management (`TLSCertificatesRequiresV4`, `write_certificates`, etc.)

### content-cache-backends-config
- **`state.py`**: Removed `hostname`, `path`, `backends_path` fields from `Configuration`
- **`config.yaml`**: Removed `hostname`, `path`, `backends-path` config options

### Documentation
- `charm-design.md`: Updated to describe port-based isolation; updated nginx config examples; replaced TLS termination section with note that TLS is handled upstream
- `tutorial/deploy-content-cache.md`: Removed `hostname=` from config command; updated curl test to use port 8080
- `how-to/enable-https.md`: Rewritten as 'connect to HTTPS backends' guide
- `reference/components.md`: Removed stale compatibility warning

## Test coverage

- 12 new TDD tests written (RED) before implementation (GREEN)
- All 93 unit tests pass (48 content-cache + 45 content-cache-backends-config)


docs/explanation/charm-design.md
```diff
@@ -46,42 +46,66 @@ archives), including use cases such as:
...
+
+This means each configured backend is reachable at a distinct port on the content-cache
+unit's IP address:
+
+```
+http://<unit-ip>:8080   →  backends from relation 1
+http://<unit-ip>:8081   →  backends from relation 2
+```
+
+An ingress component (such as haproxy with the `ingress-configurator` charm) is expected
```
erinecon: Either
```suggestion
An ingress component (such as HAProxy with the `ingress-configurator` charm) is expected
```
or
```suggestion
An ingress component (such as `haproxy` with the `ingress-configurator` charm) is expected
```
Really depends on whether you're talking about HAProxy as a workload or `haproxy` the charm (in which case you should code-format it like `` `haproxy` ``)

docs/explanation/charm-design.md
```diff
@@ -173,32 +196,17 @@ level, not the background health check level.
...
-silently serve unencrypted traffic because a certificate is delayed.
+Backends are always addressed directly by IP address over the protocol specified by the
+`protocol` configuration option (`http` or `https`).
 
-If no `certificates` relation is present, the charm does not add a `listen 443 ssl` directive
-to the nginx server block. nginx then falls back to its default behavior of listening on port 80,
-serving all traffic over plain HTTP with no TLS.
+When `protocol` is set to `https`, nginx connects to the backend over TLS. The charm does
+not manage TLS certificates for the incoming (listening) side. TLS termination for
+incoming client traffic is expected to be handled by an upstream ingress (such as haproxy with
```
erinecon: Same comment here about HAProxy or `` `haproxy` ``

docs/how-to/enable-https.md
```diff
@@ -1,19 +1,27 @@
...
 ```bash
-juju deploy self-signed-certificates cert
-juju integrate cert cache
+juju config backends protocol=https
 ```
 
-After the charm are in active status, cURL can be used to test the charm. Note the `-k` since self-signed certificates are used.
+When `protocol=https`, nginx connects to the backend IP addresses over TLS on port 443.
+
+## Skipping SSL certificate verification
```
erinecon:
```suggestion
## Skip SSL certificate verification
```

docs/how-to/enable-https.md
```diff
@@ -1,19 +1,27 @@
...
 ```bash
-curl http://<IP of the juju machine> -H "Host: <hostname in config>" -k
+juju config backends healthcheck-ssl-verify=false
 ```
+
+## TLS termination for incoming traffic
+
+The Content Cache charm does not terminate TLS for incoming client requests.
+Client-facing TLS termination is expected to be handled by an upstream ingress component,
+such as haproxy configured with the `ingress-configurator` charm.
```
erinecon: Same comment here about HAProxy or `` `haproxy` ``

docs/how-to/enable-https.md
erinecon, line 1: Do you think that the short name of this guide should be updated in `docs/how-to/index.md`? Right now it's still displayed as `Enable HTTPS`, which might be fine, but also wanted to point it out in case we should update to something like `Connect to HTTPS backends` :)

Image

erinecon (commented):
Nits only on the documentation! Thanks for updating it alongside the code :)

content-cache/src/charm.py
```diff
@@ -199,6 +145,29 @@ def _get_config_and_update_status(self) -> NginxConfig | None:
...
+        key = str(relation_id)
+        port_map: dict[str, int] = self._stored.port_map  # type: ignore[assignment]
+        if key not in port_map:
+            used_ports = set(port_map.values())
+            for offset in range(NGINX_PORT_RANGE_SIZE):
+                candidate = NGINX_PORT_RANGE_START + offset
+                if candidate not in used_ports:
+                    port_map[key] = candidate
+                    break
+        return port_map[key]
```
yhaliaw, line 186: The previous `_nginx_initialize` registered ports with Juju via `self.unit.set_ports(80, 443)`. That call was removed without a replacement for the dynamically allocated ports. nginx now listens on ports 8080–8279, but none of those ports are ever passed to `unit.set_ports` or `unit.open_port`. On an exposed unit, Juju’s firewall will not open rules for these ports and the cache service becomes unreachable after `juju expose`.

Consider calling `self.unit.set_ports(*port_map.values())` with the current allocation at the end of `_load_nginx_config`, and ensuring ports are de-registered when relations break. If opening these ports is intentionally delegated to an upstream component, document that decision explicitly.

*🤖 AI-assisted*

yhaliaw (commented):
Major changes needed

---
🤝 Human review with AI assistance.

content-cache/src/charm.py
florentianayuwono: thanks for pointing this out! i have added the logic to call it when the unit is active and also when relation is broken

content-cache/src/nginx_manager.py
```diff
@@ -376,74 +372,63 @@ def _create_status_page_config() -> None:
...
             ),
-            nginx.Key("error_log", _get_error_log_path(host, instance_name)),
+            nginx.Key("error_log", _get_error_log_path(identifier, instance_name)),
         )
 
-        if certificate_path is not None:
-            server_config.add(nginx.Key("listen", "443 ssl"))
-            server_config.add(nginx.Key("ssl_certificate", str(certificate_path)))
-            server_config.add(nginx.Key("ssl_certificate_key", str(certificate_path)))
+        upstream = str(uuid.uuid4())
```
weiiwang01: Using `uuid.uuid4` will cause the generated configuration to be different (in exact bytes) even with the same input, which might cause unnecessary reloads.

I think using `backend-{relation-id}` to replace the uuid could make it more stable.

content-cache/src/charm.py
```diff
@@ -199,6 +148,34 @@ def _get_config_and_update_status(self) -> NginxConfig | None:
         self.unit.status = ops.MaintenanceStatus(RECEIVED_NGINX_CONFIG_MESSAGE)
         return nginx_config
 
+    def _get_port_for_relation(self, relation_id: int) -> int:
```
weiiwang01, line 154: This port allocation algorithm may cause port reuse when one relation is removed and a new relation is added, and the new relation will use the old relation's port.

Port reuse is inevitable, but I think we should try to mitigate the issue, as the ingress could have some delay for relation change, so we should maximize the time interval on port number reuse on relation change.

My purpose is to always incrementally the port number on allocation, until it rolls over (like how pid is allocated in linux).

content-cache/src/charm.py
```diff
@@ -41,12 +29,16 @@
 WAIT_FOR_CONFIG_MESSAGE = "Waiting for integration with config charm"
 NGINX_NOT_READY_MESSAGE = "Nginx is not ready"
 RECEIVED_NGINX_CONFIG_MESSAGE = "Received nginx configuration"
-WAIT_FOR_TLS_CERT_MESSAGE = "Waiting for TLS certificates requested"
+
+NGINX_PORT_RANGE_START = 8080
+NGINX_PORT_RANGE_SIZE = 200
```
weiiwang01: I think starting with 8080 might be too low and can cause conflicts with other services. We should consider the 20000+ or 30000+ range.

docs/explanation/charm-design.md
```diff
@@ -46,42 +46,66 @@ archives), including use cases such as:
...
+relation always receives the same port for the lifetime of that relation, stored via Juju's
+`StoredState`.
+
+This means each configured backend is reachable at a distinct port on the content-cache
+unit's IP address:
+
+```
+http://<unit-ip>:8080   →  backends from relation 1
+http://<unit-ip>:8081   →  backends from relation 2
+```
```
erinecon: Nit: My understanding is that the code block shows how the allocation works. IMO it might be better to use a bulleted list instead, maybe something like:
```suggestion
This means each configured backend is reachable at a distinct port on the content-cache
unit's IP address:

- `http://<unit-ip>:8080` contains the backends for the first `cache-config` relation
- `http://<unit-ip>:8081` contains the backends for the second `cache-config` relation
```

docs/how-to/enable-https.md
```diff
@@ -1,19 +1,27 @@
...
+juju config backends protocol=https
 ```
 
-After the charm are in active status, cURL can be used to test the charm. Note the `-k` since self-signed certificates are used.
+When `protocol=https`, nginx connects to the backend IP addresses over TLS on port 443.
+
+## Skip SSL certificate verification
+
+If the backends use self-signed certificates, you must disable SSL verification for the
+healthcheck probes, otherwise all backends will be marked as down:
```
erinecon:
```suggestion
If the backends use self-signed certificates, you must disable SSL verification for the
healthcheck probes, or all backends will be marked as down. To disable SSL verification, run:
```

erinecon (commented):
Spotted a few more nits, otherwise the documentation LGTM 👍 

(No official approval from me since there are significant code changes in this PR. If you want an approval from me, feel free to ping me when the PR gets a technical approval)

content-cache/src/nginx_manager.py
florentianayuwono: i replaced it with backend-{identifier} with identifier = the allocated port, since the port is unique per relation and it's much simpler to get in this function

content-cache/src/charm.py
florentianayuwono: good suggestion, done!

content-cache/src/charm.py
florentianayuwono: starting from 30000 now

docs/explanation/charm-design.md
```diff
@@ -46,42 +46,69 @@ archives), including use cases such as:
...
+directive (set at the server block level) ties this location to its dedicated cache zone.
+
+## Port allocation
+
+The charm allocates a unique TCP port to each `cache-config` relation. Ports are assigned
+from a fixed range starting at `30000` and are stable across charm restarts. The same
+relation always receives the same port for the lifetime of that relation, stored via Juju's
+`StoredState`.
+
+Ports are allocated monotonically — like Linux PIDs — so that when a relation is removed
```
florentianayuwono:
```suggestion
Ports are allocated monotonically, so that when a relation is removed
```

weiiwang01 (approved):

PR #137, opened 2026-07-15
chore(docs): add Charmhub and Snapcraft listing links

erinecon:


erinecon:
Duplicated PR, I manually opened it after charmkeeper failed to. And now charmkeeper opened it :joy:

PR #138, opened 2026-07-16
feat: return cache-backends and use URL-format backends

florentianayuwono:
## Summary

- Replaces bare IP addresses + separate \`protocol\` config with full URL-format backends (\`http://ip:port\` or \`https://ip:port\`) in the \`content-cache-backends-config\` charm
- Removes the \`protocol\` option from \`content-cache-backends-config\`
- After nginx becomes active, the content-cache charm writes \`cache-backends\` to the \`cache-config\` relation unit databag as a JSON list of URLs, so ingress components can discover the address automatically
- \`cache-backends\` is cleared when the relation is removed

## Changes

- \`content-cache-backends-config\`: removed Protocol enum and protocol config; backends now accepts URLs
- \`content-cache/src/state.py\`: removed Protocol enum; backends are AnyHttpUrl; added get_cache_backends_urls() and _get_listen_protocol()
- \`content-cache/src/nginx_manager.py\`: derives scheme and port from URL attributes
- \`content-cache/src/charm.py\`: writes cache-backends to relation data after active, clears on relation-broken
- Integration tests, unit tests, and docs updated throughout

## Test Plan

- [x] Unit tests pass: tox -e unit in both charms
- [x] Lint passes: tox -e lint in both charms
- [x] Integration tests pass (requires live Juju environment)


docs/explanation/charm-design.md
```diff
@@ -203,10 +203,29 @@ nginx does not serve stale content beyond its TTL by default.
...
+(such as haproxy with the `ingress-configurator` charm).
+
+## Cache-backends published address
+
+After nginx is configured and active, the content-cache charm writes a `cache-backends` field
+to the `cache-config` relation data. This field contains a JSON list of URLs in the form
+`http://<unit-bind-ip>:<allocated-port>`, representing the address at which this unit is
+listening for the relation.
+
+An ingress component can read this value to replace its haproxy backend address with
```
erinecon:
```suggestion
An ingress component can read this value to replace its HAProxy-backend address with
```

docs/explanation/charm-design.md
```diff
@@ -203,10 +203,29 @@ nginx does not serve stale content beyond its TTL by default.
...
+the content-cache unit address.
+
+The field is updated if the port changes and cleared when the relation is removed.
```
erinecon: I'm interpreting the first part of this sentence as, the `cache-backends` field is *automatically* updated if the port changes. Is that correct?

Nit, the sentence is a bit difficult to parse, what do you think about splitting the behaviors into dedicated sentences like:
```suggestion
The field is updated if the port changes. When the relation is removed, the field is cleared.
```

erinecon (commented):
Nits only for the documentation changes! Thanks for updating the documentation alongside the code :)

content-cache/src/charm.py
```diff
@@ -127,6 +130,12 @@ def _load_nginx_config(self) -> None:
         self._update_status_with_nginx()
         if isinstance(self.unit.status, ops.ActiveStatus):
             self.unit.status = ops.ActiveStatus(status_message)
+            for rel_id, (port, _) in ported_config.items():
+                rel = self.model.get_relation(CACHE_CONFIG_INTEGRATION_NAME, rel_id)
+                if rel is None:
+                    continue
+                urls = get_cache_backends_urls(self, rel, port)
+                rel.data[self.unit]["cache-backends"] = json.dumps(urls)
```
yhaliaw: `cache-backends` is written only when the unit reaches `ActiveStatus`. If a later hook run fails validation and moves the unit to `BlockedStatus`, the previously published URL is left stale on the relation databag.

Ingress could keep routing to a backend whose configuration is now rejected, rather than seeing a cleared value as it does on the relation-broken path (where `cache-backends` is explicitly set to `""`). This is an asymmetry that may mislead operators and consuming charms.

Consider clearing `rel.data[self.unit]["cache-backends"]` to `""` on the error path in `_get_config_and_update_status` or at the top of `_load_nginx_config`, or document explicitly that stale values persist until relation-broken.

*🤖 AI-assisted*

content-cache/src/nginx_manager.py
```diff
@@ -462,16 +461,19 @@ def _get_upstream_healthchecks_worker(upstream: str, config: LocationConfig) ->
     Returns:
         A string with the lua script for the healthcheck workers.
     """
+    first_url = config.backends[0]
+    scheme = first_url.scheme
+    port = int(first_url.port) if first_url.port else (443 if scheme == "https" else 80)
```
yhaliaw: `port` is derived solely from `config.backends[0]` and then passed as the global `port` argument to `hc.spawn_checker`. That argument overrides the port used to health-check **every** peer in the upstream block.

Backends share a scheme (enforced by the validator), but they may legitimately use different ports — e.g. `http://a:8080` and `http://b:9090`. In that case every peer beyond the first is health-checked on the wrong port, producing false negatives and incorrectly marking healthy backends as down without any log indication of the mismatch.

Omit the global `port` argument so the healthcheck library uses each peer's own URL-derived port, or generate per-peer `hc.spawn_checker` calls instead.

*🤖 AI-assisted*

content-cache/src/nginx_manager.py
```diff
@@ -497,8 +499,9 @@ def _get_location_config_keys(config: LocationConfig, upstream: str) -> tuple[ng
     Returns:
         The nginx.Key for the Location configuration.
     """
+    scheme = config.backends[0].scheme
     keys = [
-        nginx.Key("proxy_pass", f"{config.protocol.value}://{upstream}/"),
+        nginx.Key("proxy_pass", f"{scheme}://{upstream}/"),
```
yhaliaw, line 502: No `proxy_set_header Host` directive is set in `_get_location_config_keys`, and the healthcheck `http_req` string carries no `Host:` header either.

With `proxy_pass http://{upstream}/` pointing at a named upstream whose name is a randomly generated UUID, nginx defaults `Host` to `$proxy_host` — that UUID string. Any origin performing virtual-host routing or `Host` header validation will misbehave or reject every request. This is likely to fail silently in tests that use a simple HTTP echo server but break against any real origin.

If the intent for this story is to leave `Host` unset, document it clearly and track it for Story 4. Otherwise add `nginx.Key("proxy_set_header", 'Host "$host"')` (or a configurable backend host) to `_get_location_config_keys`, and a `Host:` line to the `http_req` string in the healthcheck worker.

*🤖 AI-assisted*

content-cache/src/nginx_manager.py
yhaliaw: The `else` branch of this expression is dead code. `pydantic.AnyHttpUrl` always populates `.port` with the scheme default (80 for `http`, 443 for `https`), so `first_url.port` is never falsy after pydantic validation.

The branch is harmless but may confuse future readers into thinking the fallback is needed.

```suggestion
    port = int(first_url.port)
```

*🤖 AI-assisted*

content-cache/src/state.py
```diff
@@ -20,7 +19,6 @@
 CERTIFICATE_INTEGRATION_NAME = "certificates"
```
yhaliaw: `CERTIFICATE_INTEGRATION_NAME = "certificates"` is no longer referenced anywhere in `src/` after TLS handling was removed from `charm.py`.

Dead constants accumulate confusion about what the charm actively supports. Remove it, or if TLS integration is planned for Story 4 in the same file, retain it with an explicit comment marking it as reserved.

*🤖 AI-assisted*

content-cache-backends-config/src/state.py
```diff
@@ -160,15 +112,13 @@ class Configuration(pydantic.BaseModel):
...
-        protocol: The protocol to request the backends with. Can be http or https.
+        backends: The backends for this configuration as full URLs.
         fail_timeout: The time to wait before using a backend after failure.
         proxy_cache_valid: The cache valid duration.
         healthcheck: The healthcheck configuration.
     """
 
-    backends: tuple[pydantic.IPvAnyAddress, ...]
-    protocol: Protocol
+    backends: tuple[pydantic.AnyHttpUrl, ...]
```
yhaliaw, line 121: `Configuration` accepts mixed-scheme backends (e.g. one `http://` URL and one `https://` URL) without error. The equivalent check exists on `LocationConfig` in `content-cache` (via `validate_backends_scheme`) but is absent here.

When an operator supplies mixed-scheme backends the failure surfaces on the consumer charm (`content-cache`) rather than at the source, making the error harder to trace and requiring the operator to investigate the wrong application.

Add the same `validate_backends_scheme` field validator to `Configuration` so the config charm rejects invalid input immediately and the error message points to the correct application.

*🤖 AI-assisted*

content-cache/tests/unit/test_charm.py
```diff
@@ -277,3 +274,49 @@ def test_port_stable_for_same_relation(harness: Harness, charm: ContentCacheChar
...
+        CACHE_CONFIG_INTEGRATION_NAME,
+        remote_app="config",
+        app_data=SAMPLE_INTEGRATION_DATA,
+    )
+
+    assert charm.unit.status == ops.ActiveStatus()
+
+    harness.remove_relation(relation_id)
+
+    assert charm.unit.status == ops.BlockedStatus(WAIT_FOR_CONFIG_MESSAGE)
```
yhaliaw, line 322: `_get_upstream_healthchecks_worker` derives the healthcheck port from `config.backends[0]` only (`port = int(first_url.port) if first_url.port else …`). With backends on different ports — e.g. `http://a:8080` and `http://b:9090` — every peer beyond the first is health-checked on the wrong port.

No test covers this mixed-port scenario to pin down the first-backend-wins behaviour or catch a future regression.

*🤖 AI-assisted*

yhaliaw (commented):
Major changes needed

---
🤝 Human review with AI assistance.

docs/explanation/charm-design.md
florentianayuwono: yep it's automatically updated

content-cache/src/nginx_manager.py
florentianayuwono: hostname based routing is intentionally removed in previous PR, so this is intentional. we won't route based on hostname anymore.

content-cache/src/state.py
```diff
@@ -159,19 +145,41 @@ class LocationConfig(pydantic.BaseModel):
...
+        Returns:
+            The value after validation.
+        """
+        if len(value) > 1:
+            schemes = {url.scheme for url in value}
+            if len(schemes) > 1:
+                raise ValueError(
+                    f"All backends must share the same scheme; found mixed schemes: {schemes}"
+                )
+        return value
```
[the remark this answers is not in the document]
florentianayuwono: healthchecks now doesn't assume a single port. ports don't need to match

docs/explanation/charm-design.md
```diff
@@ -203,10 +203,29 @@ nginx does not serve stale content beyond its TTL by default.
...
-incoming client traffic is expected to be handled by an upstream ingress (such as haproxy with
-the `ingress-configurator` charm).
+For example, to proxy to an HTTPS backend:
+
+```bash
+juju config backends backends=https://185.125.90.20:443
+```
+
+When the URL scheme is `https`, nginx connects to the backend over TLS.
+All backends in a single relation must use the same scheme.
```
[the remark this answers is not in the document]
florentianayuwono: backends do not need tos hare the same port.

content-cache/src/nginx_manager.py
```diff
@@ -442,12 +441,12 @@ def _get_upstream_config_keys(config: LocationConfig) -> tuple[nginx.Key, ...]:
...
-        port = 443
     keys = [
-        nginx.Key("server", f"{ip}:{port} fail_timeout={config.fail_timeout}")
-        for ip in config.backends
+        nginx.Key(
+            "server",
+            f"{url.host}:{url.port} fail_timeout={config.fail_timeout}",
+        )
+        for url in config.backends
     ]
```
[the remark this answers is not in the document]
florentianayuwono: the AnyHttpUrl guarantees that the port is always set (with the default fallback) so no code changes needed

content-cache/tests/unit/test_charm.py
florentianayuwono: the behavior has been fixed to allow for different ports. healthcheck tests have also been added!

docs/explanation/charm-design.md
```diff
@@ -203,10 +203,29 @@ nginx does not serve stale content beyond its TTL by default.
...
-Backends are always addressed directly by IP address over the protocol specified by the
-`protocol` configuration option (`http` or `https`).
+Backends are specified as full URLs in the form `<http|https>://<ip>:<port>`.
+The protocol and port are encoded directly in each backend URL.
 
-When `protocol` is set to `https`, nginx connects to the backend over TLS. The charm does
-not manage TLS certificates for the incoming (listening) side. TLS termination for
-incoming client traffic is expected to be handled by an upstream ingress (such as `haproxy` with
-the `ingress-configurator` charm).
+For example, to proxy to an HTTPS backend:
```
erinecon:
```suggestion
For example, to proxy to an HTTPS backend, run:
```
or

```suggestion
For example, proxy to an HTTPS backend using:
```

docs/how-to/enable-https.md
```diff
@@ -2,14 +2,14 @@
 
 # How to connect to HTTPS backends
 
-The Content Cache charm can proxy to backends over HTTPS by setting the `protocol` option
+The Content Cache charm can proxy to backends over HTTPS by using HTTPS URLs in the `backends` option
 on the `content-cache-backends-config` charm.
```
erinecon, line 6:
```suggestion
on the `content-cache-backends-config` charm.
The following example sets the backend to `https://10.10.1.1:443`:
```

docs/how-to/enable-https.md
```diff
@@ -2,14 +2,14 @@
...
 
 ```bash
-juju config backends protocol=https
+juju config backends backends=https://10.10.1.1:443
 ```
 
-When `protocol=https`, nginx connects to the backend IP addresses over TLS on port 443.
+When the URL scheme is `https`, nginx connects to the backend over TLS on the specified port.
```
erinecon:
```suggestion
When the URL scheme is `https`, nginx connects to the backend over TLS on the specified port
(which is `443` in the example above).
```

docs/how-to/enable-https.md
```diff
@@ -2,14 +2,14 @@
 
 # How to connect to HTTPS backends
 
-The Content Cache charm can proxy to backends over HTTPS by setting the `protocol` option
+The Content Cache charm can proxy to backends over HTTPS by using HTTPS URLs in the `backends` option
 on the `content-cache-backends-config` charm.
 
 ```bash
-juju config backends protocol=https
+juju config backends backends=https://10.10.1.1:443
```
erinecon: I'm a bit concerned that users won't understand that `backends` is the `content-cache-backends-config` charm since we don't explicitly say anything about it here. I think it would be clearer if this example said:
```suggestion
juju config content-cache-backends-config backends=https://10.10.1.1:443
```
Otherwise we should explicitly state the mapping between the charm and `backends` in the example commands

erinecon (commented):
Nits only on my end -- thanks for updating the documentation 👍 

(No official approval on my end since there are significant code changes in this PR. If you'd like an approval from me, feel free to ping me after the PR has a technical approval)

yhaliaw (approved):

weiiwang01 (approved):

PR #146, opened 2026-08-25
feat(isd-6382): add backend-hostname config for HTTPS backend verification

florentianayuwono:
## Summary

Adds `backend-hostname` to the `cache-config` relation and `content-cache-backends-config` charm to support HTTPS backend verification.

- **`backend-hostname`**: Sets `proxy_ssl_name` and the HTTP `Host` header when proxying to the backend. **Mandatory for HTTPS backends.**
- When set, nginx uses the combined CA bundle (`/etc/nginx/certs/ca-bundle.pem`) as `proxy_ssl_trusted_certificate`. This bundle always contains the Ubuntu system CAs (covering public certs such as Let's Encrypt), plus any custom CA certs provided via the `receive-ca-cert` relation — enabling both public and private/self-signed backend CAs.

When `backend-hostname` is set for an HTTPS backend, nginx emits:
```nginx
proxy_set_header Host <backend_hostname>;
proxy_ssl_name <backend_hostname>;
proxy_ssl_server_name on;
proxy_ssl_verify on;
proxy_ssl_verify_depth 10;
proxy_ssl_trusted_certificate /etc/nginx/certs/ca-bundle.pem;
```

### `content-cache-backends-config`
- Added `backend-hostname` config option (example: `radosgw.ps7.canonical.com`)
- Field sent to `content-cache` via the `cache-config` relation

### `content-cache`
- **`ca_certs.py`**: `write_ca_bundle` always prepends Ubuntu system CAs before any operator-supplied certs, ensuring the bundle is always populated. `SYSTEM_CA_BUNDLE_PATH` constant defined here.
- **`charm.py`**: `_rebuild_ca_bundle` is called at charm startup (and on every `receive-ca-cert` change/broken event) to keep the combined bundle up to date even without any cert-transfer relations.
- **`nginx_manager.py`**: Emits 6 nginx SSL directives when `backend_hostname` is set; `proxy_ssl_trusted_certificate` points to the combined bundle.
- **`state.py`** (`LocationConfig`): `backend_hostname` field with model validator requiring it for HTTPS backends.

### Removed
- `backend-ca-fingerprint` (previously proposed) — not needed; the combined bundle approach covers both public and private CAs without fingerprint lookup.
- `_ensure_ca_bundle_ready` gate — the bundle always exists (system CAs are always included).
- Unconditional `proxy_ssl_verify on` (previously caused 502 when no backend-hostname was set).

## Test Plan
- [x] Unit tests: 71 pass in `content-cache/`, 42 pass in `content-cache-backends-config/`
- [x] Lint + static: clean in both charms
- [x] Integration tests: `test_healthchecks_ssl_verify` (two parametrised cases: trusted cert → 200, untrusted → 502), `test_certificate_transfer_full_lifecycle` (HTTPS lifecycle with cert-transfer)


content-cache-backends-config/src/state.py
```diff
@@ -24,6 +24,26 @@
...
+BACKEND_HOSTNAME_CONFIG_NAME = "backend-hostname"
+BACKEND_CA_FINGERPRINT_CONFIG_NAME = "backend-ca-fingerprint"
+
+
+def _validate_hostname_value(value: str) -> str:
+    """Validate the value as a hostname."""
+    if len(value) > 255:
+        raise ValueError("Hostname cannot be longer than 255")
+
+    valid_segment = re.compile(r"(?!-)[A-Z\d-]{1,63}(?<!-)$", re.IGNORECASE)
```
florentianayuwono: can you add code comment explain what each pattern is scanning for?

content-cache-backends-config/src/state.py
```diff
@@ -51,6 +71,27 @@ def _validate_path_value(value: str) -> str:
...
+    if not value:
+        return value
+    return _validate_hostname_value(value)
+
+
+def _validate_fingerprint_value(value: str) -> str:
+    """Validate the value as a colon-separated SHA-256 hex fingerprint."""
+    if not value:
+        return value
+    pattern = re.compile(r"^[0-9A-Fa-f]{2}(:[0-9A-Fa-f]{2}){31}$")
```
florentianayuwono: can you add code comment to explain what each pattern is looking for

content-cache-backends-config/src/state.py
```diff
@@ -213,6 +269,8 @@ def from_charm(cls, charm: ops.CharmBase) -> "Configuration":
                 fail_timeout=fail_timeout,
                 proxy_cache_valid=proxy_cache_valid,  # type: ignore
                 healthcheck=healthcheck_config,
+                backend_hostname=backend_hostname,
+                backend_ca_fingerprint=backend_ca_fingerprint,  # type: ignore
```
florentianayuwono: why is there a type: ignore here

content-cache-backends-config/config.yaml
```diff
@@ -47,3 +47,20 @@ options:
       (https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_valid).
     type: string
     default: '[]'
+  backend-hostname:
+    description: |
+      The hostname to use as the SNI and HTTP Host header when proxying to HTTPS backends.
+      Required when backends use https://. Must be a valid DNS hostname (max 255 characters,
+      alphanumeric segments separated by dots).
```
florentianayuwono: can you give example as well

content-cache/src/ca_certs.py
```diff
@@ -50,3 +55,95 @@ def get_ca_bundle_path() -> Path | None:
...
+        The matching PEM certificate, or None if no match is found.
+    """
+    normalised = fingerprint.strip().upper()
+    for pem in pem_certs:
+        pem = pem.strip()
+        if not pem:
+            continue
+        try:
+            cert_fingerprint = _cert_sha256_fingerprint(pem)
+        except Exception:  # noqa: BLE001  # pylint: disable=broad-exception-caught
```
florentianayuwono: what kind of exceptions do you foresee happening? i think it's better to catch those that are expected and then let go of those that's unforeseen, don't try to catch all

content-cache/src/ca_certs.py
```diff
@@ -50,3 +55,95 @@ def get_ca_bundle_path() -> Path | None:
...
+            continue
+        if cert_fingerprint == normalised:
+            return pem
+    return None
+
+
+def write_backend_ca_cert(identifier: str, cert_pem: str) -> Path:
```
florentianayuwono: why do we need write_backend_ca_cert and remove_backend_ca_cert when there's already write_ca_bundle that reconcile all cert

content-cache-backends-config/src/state.py
florentianayuwono: Added inline comments to the regex: `(?!-)` = segment must not start with a hyphen, `[A-Z\d-]{1,63}` = alphanumeric+hyphen, max 63 chars per segment, `(?<!-)$` = segment must not end with a hyphen.

weiiwang01 (approved):

content-cache/tox.ini
```diff
@@ -62,7 +62,8 @@ commands =
     # codespell {[vars]lib_path}
     codespell {toxinidir} --skip {toxinidir}/.git --skip {toxinidir}/.tox \
       --skip {toxinidir}/build --skip {toxinidir}/lib --skip {toxinidir}/venv \
-      --skip {toxinidir}/.mypy_cache --skip {toxinidir}/icon.svg
+      --skip {toxinidir}/.mypy_cache --skip {toxinidir}/icon.svg \
+      --skip {toxinidir}/tests/integration/scripts/certificate.pem
```
florentianayuwono: why do we skip the tox check here

content-cache-backends-config/config.yaml
```diff
@@ -47,3 +47,11 @@ options:
...
+      Example: backend-hostname=radosgw.ps7.canonical.com
```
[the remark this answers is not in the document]
florentianayuwono: Changed to `backend.example.com`.

content-cache/tests/integration/test_tls_cert.py
```diff
@@ -3,56 +3,104 @@
...
     await cache_tester.integrate_config()
     config = dict(CacheTester.BASE_CONFIG)
-    config[BACKENDS_CONFIG_NAME] = f"https://{http_ok_ip}:443"
+    config[BACKENDS_CONFIG_NAME] = f"https://{backend_ip}:443"
+    config[BACKEND_HOSTNAME_CONFIG_NAME] = "localhost"
+    config[HEALTHCHECK_PATH_CONFIG_NAME] = "/"
+    config[HEALTHCHECK_INTERVAL_CONFIG_NAME] = str(HEALTHCHECK_INTERVAL)
+    config[HEALTHCHECK_SSL_VERIFY_CONFIG_NAME] = "false"
+    config[HEALTHCHECK_VALID_STATUS_CONFIG_NAME] = "200"
+    config[PROXY_CACHE_VALID_CONFIG_NAME] = '["200 10s"]'
```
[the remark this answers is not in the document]
florentianayuwono: Set `proxy-cache-valid` to `["200 1s"]` in the lifecycle test so Phase 3 cannot be served from a cached Phase 2 response.

content-cache/src/ca_certs.py
```diff
@@ -13,40 +13,33 @@
 CA_CERTS_DIR = Path("/etc/nginx/certs")
 CA_BUNDLE_PATH = CA_CERTS_DIR / "ca-bundle.pem"
 
+# Ubuntu system CA bundle — always included as the base trust store so that
+# publicly trusted certificates (e.g. Let's Encrypt / ISRG Root X1) are
+# verified without manual configuration.
+SYSTEM_CA_BUNDLE_PATH = Path("/etc/ssl/certs/ca-certificates.crt")
```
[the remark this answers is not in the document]
florentianayuwono: Added `ca-certificates` to `charm-system-packages` in `content-cache/charmcraft.yaml`.

content-cache/tox.ini
florentianayuwono: `certificate.pem` contains base64-encoded binary data that codespell misidentifies as misspellings (e.g. common letter sequences in the base64 alphabet). Skipping it avoids those false positives.

content-cache-backends-config/src/state.py
```diff
@@ -116,12 +145,16 @@ class Configuration(pydantic.BaseModel):
...
+        backend_hostname: Hostname used for backend SNI and Host header.
     """
 
     backends: tuple[pydantic.AnyHttpUrl, ...]
     fail_timeout: typing.Annotated[str, pydantic.StringConstraints(min_length=1)]
     proxy_cache_valid: tuple[str, ...]
     healthcheck: HealthcheckConfig
+    backend_hostname: typing.Annotated[
+        str, pydantic.AfterValidator(_validate_optional_hostname_value)
+    ] = ""
```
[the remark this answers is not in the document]
florentianayuwono: Added a `model_validator` on `Configuration` that raises `ValueError` when any backend uses `https://` and `backend_hostname` is empty. Also added a unit test `test_https_backends_without_hostname_raises` for this case.

content-cache/src/ca_certs.py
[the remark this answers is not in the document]
florentianayuwono: `charm-system-packages` is rejected by the current charmcraft version with a validation error (`extra_forbidden`). `ca-certificates` is part of the default Ubuntu install, so `/etc/ssl/certs/ca-certificates.crt` is already present on Juju machines. Removed the stanza.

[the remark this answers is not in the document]
florentianayuwono: Updated approach: added `ca-certificates` to the `apt install` call in `nginx_manager.initialize()` so it is installed at charm startup alongside nginx. `charm-system-packages` and `stage-packages` don't cover the system path `/etc/ssl/certs/ca-certificates.crt` used at runtime.

